PR #991, opened 2022-08-10
Automation test  cs 1480

LanaDIO:


apps/tenant-management-webapp-e2e/src/integration/events/events.step.ts
```diff
@@ -171,7 +171,221 @@ When('the user clicks eye icon of {string} under Core streams', function (stream
...
+    }
+  });
+});
+
+Then(
+  'the user views the stream details of {string}, {string}, {string}, {string}',
+  function (streamName, description, event, role) {
+    eventsObj.streamDetails().should('contain', streamName);
+    eventsObj.streamDetails().should('contain', description);
+    eventsObj.streamDetails().should('contain', '"publicSubscribe": false');
```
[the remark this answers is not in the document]
LanaDIO: added

apps/tenant-management-webapp-e2e/src/integration/events/events.page.ts
```diff
@@ -127,11 +127,109 @@ class eventsPage {
...
+  }
+
+  streamDeleteBtn(name) {
+    return cy.xpath(
+      `//*[@data-testid="stream-name" and contains(text(), "${name}")]/following-sibling::td//*[@data-testid="delete-stream"]`
+    );
+  }
+
+  streamModalEventChip(name) {
+    return cy.get(`[content="${name}"]`);
```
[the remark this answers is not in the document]
LanaDIO: I tried, can not come up with anything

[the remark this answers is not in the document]
howard-zhang: Updated the element to be goa-chip for all chips. In the step to use shadow().get() to find the name under goa-chip elements, so that the name will be searched only within chips, not other UI elements.

howard-zhang (changes requested):

apps/tenant-management-webapp-e2e/src/integration/events.feature
```diff
@@ -102,4 +102,31 @@ Feature: Events
...
\ No newline at end of file
+    Then the user views the details of "PDF generation updates" under Core streams
+
+  @TEST_CS-1480 @REQ_CS-1352 @regression
+  Scenario: As a tenant admin, I can add, edit and delete an event stream
+    Given a tenant admin user is on event streams page
+    # Add a stream
+    When the user clicks Add stream button
+    Then the user views Add stream modal
+    When the user enters "autotest-addEditDelete-stream", "autotest-stream-desc", "status-service:application-healthy, file-service:file-deleted", "auto-test-role1, auto-test-role2" in stream modal
```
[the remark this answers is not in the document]
LanaDIO: changed to public

apps/tenant-management-webapp-e2e/src/integration/events/events.page.ts
```diff
@@ -127,11 +127,115 @@ class eventsPage {
     );
   }
 
-  streamDetails(streamName) {
+  streamDetailsTable(streamName) {
```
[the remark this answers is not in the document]
LanaDIO: because I am using it in this object
streamDetails() {
    return cy.get('[data-testid="stream-details"]');
  }

apps/tenant-management-webapp-e2e/src/integration/events/events.step.ts
```diff
@@ -171,7 +171,260 @@ When('the user clicks eye icon of {string} under Core streams', function (stream
...
+      .streamModalPublicCheckbox()
+      .invoke('attr', 'class')
+      .then((classAttr) => {
+        if (classAttr?.includes('-selected')) {
+          cy.log('Make stream public is already checked off.');
+        } else {
+          eventsObj.streamModalPublicCheckbox().click();
+        }
+      });
+  } else {
```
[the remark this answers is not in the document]
LanaDIO: the code for unckecking already here look below

apps/tenant-management-webapp-e2e/src/integration/events/events.step.ts
```diff
@@ -171,7 +171,260 @@ When('the user clicks eye icon of {string} under Core streams', function (stream
...
+        .streamModalPublicCheckbox()
+        .invoke('attr', 'class')
+        .then((classAttr) => {
+          if (classAttr?.includes('-selected')) {
+            cy.log('Make stream public is already checked off.');
+          } else {
+            eventsObj.streamModalPublicCheckbox().click();
+          }
+        });
+    } else {
```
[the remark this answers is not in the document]
LanaDIO: this is new entry the public is not checked by default, there is no need for this code in this step.

apps/tenant-management-webapp-e2e/src/integration/events/events.page.ts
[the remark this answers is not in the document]
howard-zhang: Why do we need 2 page objects for stream details? Looks to me streamDetails() should be renamed to streamDetailsAll() as it's a page object for all stream details on the page. streamDetailsTable() isn't a good page object name as user will have no clue what this mean and there is no table for stream details on the page. If you want to include table in the name, it should be streamTableDetails(). 

I also looked at the places that streamDetails() is being used, and I think it should be replaced by the page object with stream name parameter to make sure the stream detail is the one for the stream. For example:
eventsObj.streamDetails().should('contain', streamName);
This is to check whatever details on the page has streamName. What the code should validate is the details under the stream name contains the stream name.
eventsObj.streamDetails(streamName).should('contain', streamName);

howard-zhang (changes requested):

apps/tenant-management-webapp-e2e/src/integration/events/events.page.ts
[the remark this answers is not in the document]
LanaDIO: Will rename it to streamTableDetails()
Added streamName as part of object verification

apps/tenant-management-webapp-e2e/src/integration/events/events.step.ts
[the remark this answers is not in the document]
LanaDIO: Good point, I can remove it. The new entry is different because edit does not have a name entry, I want to keep it simple.

apps/tenant-management-webapp-e2e/src/integration/events.feature
```diff
@@ -102,4 +102,31 @@ Feature: Events
...
+    And the user clicks Save button in Stream modal
+    Then the user "views" the stream of "autotest-addEditDelete-stream", "Public"
+    When the user clicks "Eye" button of "autotest-addEditDelete-stream"
+    Then the user views the stream details of "autotest-addEditDelete-stream", "autotest-stream-desc", "status-service:application-healthy, file-service:file-deleted", "public"
+    And the user clicks "Eye-Off" button of "autotest-addEditDelete-stream"
+    # Edit a stream
+    When the user clicks "Edit" button of "autotest-addEditDelete-stream"
+    Then the user views Edit stream modal
+    When the user removes event chips of "status-service:application-healthy" in stream modal
+    And the user enters "autotest-stream-desc2", "auto-test-role3, beta-tester" in stream modal
```
howard-zhang, line 123: When the user enters "autotest-stream-desc2", "EMPTY", "auto-test-role3, beta-tester" in Edit stream modal

apps/tenant-management-webapp-e2e/src/integration/events.feature
```diff
@@ -102,4 +102,31 @@ Feature: Events
...
\ No newline at end of file
+    Then the user views the details of "PDF generation updates" under Core streams
+
+  @TEST_CS-1480 @REQ_CS-1352 @regression
+  Scenario: As a tenant admin, I can add, edit and delete an event stream
+    Given a tenant admin user is on event streams page
+    # Add a stream
+    When the user clicks Add stream button
+    Then the user views Add stream modal
+    When the user enters "autotest-addEditDelete-stream", "autotest-stream-desc", "status-service:application-healthy, file-service:file-deleted", "public" in stream modal
```
howard-zhang, line 113: When the user enters "autotest-addEditDelete-stream", "autotest-stream-desc", "status-service:application-healthy, file-service:file-deleted", "public" in Add stream modal

apps/tenant-management-webapp-e2e/src/integration/events/events.page.ts
```diff
@@ -126,10 +126,113 @@ class eventsPage {
       `//*[@data-testid="Platform-stream-table"]//tbody/tr/td[text()="${streamName}"]/following-sibling::td//button[@data-testid="toggle-stream-visibility"]`
     );
   }
+  streamTableDetails(streamName) {
```
howard-zhang, line 129: This one is a duplicated page object of streamDetails.

howard-zhang (changes requested):

apps/tenant-management-webapp-e2e/src/integration/events/events.page.ts
LanaDIO: This one object works for both tables tenant and core

apps/tenant-management-webapp-e2e/src/integration/events.feature
LanaDIO: added Add to the name and also added the case where user can enter n/a

apps/tenant-management-webapp-e2e/src/integration/events.feature
LanaDIO: Added Edit to the name of the test case and "n/a"

apps/tenant-management-webapp-e2e/src/integration/events/events.step.ts
```diff
@@ -175,3 +175,257 @@ Then('the user views the details of {string} under Core streams', function (stre
...
+    else {
+      eventsObj.streamDetails(streamName).should('contain', '"publicSubscribe": false');
+    }
+    const roles = role.split(',');
+    for (let i = 0; i < roles.length; i++) {
+      eventsObj
+        .streamDetails(streamName)
+        .invoke('text')
+        .then((roleDetails) => {
+          const subscriberRoles = roles[i].split(',')[0].trim();
```
howard-zhang, line 321: Why do we need another .split(',') here (split() has been done at line 315)? I believe roles[i].trim() should be enough as there should be no "," in the variable  already.

apps/tenant-management-webapp-e2e/src/integration/events/events.step.ts
```diff
@@ -175,3 +175,257 @@ Then('the user views the details of {string} under Core streams', function (stre
...
+
+When('the user clicks Add stream button', function () {
+  eventsObj.addStreamBtn().click();
+});
+
+Then('the user views Add stream modal', function () {
+  eventsObj.streamModalTitle().invoke('text').should('eq', 'Add stream');
+});
+
+When(
```
howard-zhang, line 199: Since we have EMPTY and n/a at different places now. We should start to add a comment on top of the step definition bout n/a and which field allows n/a. Please add a comment there.

apps/tenant-management-webapp-e2e/src/integration/events/events.step.ts
```diff
@@ -175,3 +175,257 @@ Then('the user views the details of {string} under Core streams', function (stre
...
+    default:
+      expect(button).to.be.oneOf(['Eye', 'Edit', 'Delete']);
+  }
+});
+
+Then('the user views Edit stream modal', function () {
+  eventsObj.streamModalTitle().invoke('text').should('eq', 'Edit stream');
+});
+
+Then('the user enters {string}, {string}, {string} in Edit stream modal', function (description, event, role) {
```
howard-zhang, line 363: Need a comment about n/a so that user calling this step knows which parameter(s) allow n/a and what n/a parameter does.

apps/tenant-management-webapp-e2e/src/integration/events/events.step.ts
```diff
@@ -175,3 +175,257 @@ Then('the user views the details of {string} under Core streams', function (stre
...
+      .streamModalPublicCheckbox()
+      .invoke('attr', 'class')
+      .then((classAttr) => {
+        if (classAttr?.includes('-selected')) {
+          cy.log('Make stream public is already checked off.');
+        } else {
+          eventsObj.streamModalPublicCheckbox().click();
+        }
+      });
+    if (role == 'n/a') {
```
howard-zhang, line 396: Should be "else if" instead of "if"

howard-zhang (changes requested):

apps/tenant-management-webapp-e2e/src/integration/events/events.step.ts
LanaDIO: yes it needed it there

apps/tenant-management-webapp-e2e/src/integration/events/events.step.ts
LanaDIO: Not sure which one, it gives me red underlining

howard-zhang (approved):

PR #1012, opened 2022-08-16
fix(status-service): fix algorithm for populating status bar

rstyan:


rstyan:
This PR fixes up the algorithm used to populate the health status bar.  Instead of trying to piece together status from a sparse array, we use time buckets to decide where a status goes.

PR #1124, opened 2022-09-06
add delete function to calendar

jonathanweyermann:
image
image


athena-chen-chen:
I think delete modal could be consistent with another service deletion.

solittlework (approved):

PR #1138, opened 2022-09-08
CS-1709 - don't show table of calendar to be deleted

jonathanweyermann:
image


solittlework (approved):

apps/tenant-management-webapp/src/app/pages/admin/services/calendar/calendarList.tsx
```diff
@@ -127,47 +129,18 @@ export const CalendarTableComponent: FunctionComponent<calendarTableProps> = ({
...
-                </tbody>
-              </DataTable>
-            ) : (
-              ''
-            )}
+            <div>Delete {selectedDeleteCalendar?.name}?</div>
           </div>
         }
         onDelete={() => {
+          setShowDeleteConfirmation(false);
```
smettu007, line 142: should we dispatch the call first and then close the modal after?

jonathanweyermann: its done this way elsewhere too. its asynchronous so it doesnt matter

smettu007 (approved):

PR #1146, opened 2022-09-09
CS-1691 - Fix Critical security issues pointed out by github

jonathanweyermann:
Addresses the following

image

### DIFF
--- a/apps/pdf-service/src/file.ts
+++ b/apps/pdf-service/src/file.ts
@@ -9,8 +9,9 @@ class PlatformFileService implements FileService {
 
   async typeExists(tenantId: AdspId, typeId: string): Promise<boolean> {
     try {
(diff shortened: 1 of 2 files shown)


tzuge:
Is `encodeUri` or `encodeUriComponent` appropriate here? Express validators also have sanitization functionality, so potentially we sanitize the parameter on the way in as well, but not sure if CodeQL will pick up on that.

solittlework (approved):

jonathanweyermann:
> Is encodeUri or encodeUriComponent appropriate here? Express validators also have sanitization functionality, so potentially we sanitize the parameter on the way in as well, but not sure if CodeQL will pick up on that.

using encodeUri in this case seems superfluous and unnecessary, as by itself encoding and decoding won't prevent us from passing allow characters that maybe bad, for instance a slash. We'd still have to filter out bad characters in the end, to the degree that we can get rid of them knowing what we know about the characters inside the input we are passing along.

PR #1152, opened 2022-09-12
update the pdf template

solittlework:


apps/tenant-management-webapp/src/app/pages/admin/services/pdf/templates/templates.tsx
```diff
@@ -256,9 +256,7 @@ export const PdfTemplates: FunctionComponent<PdfTemplatesProps> = ({ openAddTemp
 
                   setCurrentChannel(channel);
                 }}
-                bodyEditorHintText={
-                  "*GOA default styles are applied if the template doesn't include proper <html> opening and closing tags"
-                }
+                bodyEditorHintText={''}
```
jonathanweyermann, line 259: Do we not need this anymore on the main page?

jonathanweyermann: and if we don't, we should also remove the place on the main page where this text is being used, and maybe resize the page so it matches the footer/header page - see https://github.com/GovAlta/adsp-monorepo/pull/1151

jonathanweyermann (changes requested):
see above

apps/tenant-management-webapp/src/app/pages/admin/services/pdf/templates/templates.tsx
solittlework: Ok, I see. In this case, I am going to remove it.

jonathanweyermann (approved):

PR #1156, opened 2022-09-12
CS-1693 - High CodeQL errors

jonathanweyermann:
Email - https://stackoverflow.com/questions/42687979/email-validation-regular-expression-is-causing-catastrophic-backtracking

### DIFF
--- a/apps/status-app/src/app/pages/status.tsx
+++ b/apps/status-app/src/app/pages/status.tsx
@@ -25,6 +25,7 @@ import moment from 'moment';
 import GoaLogo from '../../assets/goa-logo.svg';
 import { GoAButton } from '@abgov/react-components';
 import { GoAForm, GoAFormItem, GoAInputEmail, GoAFormActions } from '@abgov/react-components/experimental';
+import { emailError } from '@lib/inputValidation';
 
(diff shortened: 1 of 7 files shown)


apps/status-app/src/app/pages/status.tsx
```diff
@@ -147,7 +149,7 @@ const ServiceStatusPage = (): JSX.Element => {
   };
 
   function emailErrors() {
-    if (!/^\w+([.-]?\w+)*@\w+([.-]?\w+)*(\.\w{2,3})+$/.test(email)) {
+    if (!/^\w+(?:[.-]\w+)*@\w+(?:[.-]\w+)*(?:\.\w{2,3})+$/.test(email)) {
```
rstyan, line 152: Could you export the regex in one of the files and import it into the rest? It's pretty complicated and would be simpler to maintain if it is written only in one place and reused in the others.

rstyan (dismissed):

apps/status-app/src/app/pages/status.tsx
jonathanweyermann: sure. I should probably do that

smettu007 (approved):

PR #1274, opened 2022-10-17
fix: add code to synchronize status service config and repo

rstyan:
by removing repo orphans; those without a corresponding configuration.


solittlework (approved):
The PR looks good. I leave one comment and there are a few small lint issues. But, they are trivial.

PR #1318, opened 2022-10-27
fix: change notice references to use appKey

rstyan:


solittlework (approved):
Hi Roy, thank you for the PR. Except for a tiny lint, the PR is good. For the compatibility issues, I would suggest asking Ting to see whether we can refresh the db to avoid to handle this.

PR #1329, opened 2022-11-04
feat: view configuration revisions

athena-chen-chen:
Screen Shot 2022-11-04 at 3 19 29 PM
Screen Shot 2022-11-04 at 3 16 45 PM

### DIFF
--- a/apps/tenant-management-webapp/src/app/store/configuration/sagas.ts
+++ b/apps/tenant-management-webapp/src/app/store/configuration/sagas.ts
@@ -23,6 +23,12 @@ import {
   ResetReplaceConfigurationListAction,
   resetReplaceConfigurationListSuccessAction,
   RESET_REPLACE_CONFIGURATION_LIST_ACTION,
+  FETCH_CONFIGURATION_REVISIONS_ACTION,
+  FetchConfigurationRevisionsAction,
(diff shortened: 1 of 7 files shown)


apps/tenant-management-webapp/src/app/store/configuration/sagas.ts
```diff
@@ -126,7 +132,59 @@ export function* fetchConfigurations(action: FetchConfigurationsAction): SagaIte
...
+  const token: string = yield call(getAccessToken);
+  const service = action.service.split(':');
+  const url = `${configBaseUrl}/configuration/v2/configuration/${service[0]}/${service[1]}/active`;
+  if (configBaseUrl && token) {
+    try {
+      const { data } = yield call(axios.get, url, { headers: { Authorization: `Bearer ${token}` } });
+
+      yield put(getConfigurationActiveSuccess(data, action.service));
+    } catch (err) {
+      yield put(getConfigurationActiveSuccess(null, action.service));
```
solittlework, line 184: It looks like we are kind of making the exception too slience. We can improve it later.

solittlework (approved):

apps/tenant-management-webapp/src/app/store/configuration/sagas.ts
athena-chen-chen: Exception means no active revision.

PR #1330, opened 2022-11-04
Add the try run endpoint for script service

solittlework:


apps/script-service/Controllers/RunScriptRequest.cs
```diff
@@ -16,4 +17,8 @@ public class RunScriptRequest
 
   [JsonPropertyName("correlationId")]
   public string? CorrelationId { get; set; }
+
```
tzuge, line 20: You should reuse the same request body POCO in separate endpoints unless they're actually identical. Define a new POC instead for the request values.

Also the test run shouldn't require a full 'definition'. The consumer of the API should be able to provide just the script as a string in addition to the input values.

apps/script-service/Controllers/ScriptController.cs
```diff
@@ -107,4 +107,38 @@ public async Task<IEnumerable<object>> RunScript(string? script, [FromBody] RunS
...
+    {
+      throw new RequestArgumentException("missing inputs attribute in the body");
+    }
+    var user = HttpContext.GetAdspUser();
+
+    Func<Task<string>> getToken = () => Task.FromResult(HttpContext.Request.Headers[HeaderNames.Authorization].First()[TOKEN_INDEX..]);
+    using (HttpContext.Benchmark("run-script-time"))
+    {
+
+      var outputs = await _luaService.RunScript(
```
tzuge, line 137: You probably need to push the test script concept down to the LuaScriptService as well, since many of the input arguments from the regular run are not applicable. Also the test run should not signal the script run events.

tzuge (changes requested):

apps/script-service/Controllers/RunScriptRequest.cs
solittlework: Hi @tzuge, thank you for your review. I've created another request schema with minimum attributes in it.

apps/script-service/Controllers/ScriptController.cs
solittlework: Hi @tzuge, I created another function - TestScript for this endpoint. Please help me review it again. Thanks!

apps/script-service/Controllers/RunScriptRequest.cs
```diff
@@ -3,6 +3,7 @@
 using System.Diagnostics.CodeAnalysis;
 using System.Text.Json.Serialization;
 using Adsp.Sdk.Util;
+using Adsp.Platform.ScriptService.Model;
```
tzuge, line 6: this is no longer needed?

apps/script-service/Services/LuaScriptService.cs
```diff
@@ -23,6 +23,37 @@ public LuaScriptService(ILogger<LuaScriptService> logger, IServiceDirectory dire
...
+    string script,
+    AdspId tenantId
+  )
+  {
+
+    _logger.LogDebug("Testing script for tenant {TenantId}...", tenantId);
+    try
+    {
+      using var lua = new Lua();
+
```
tzuge, line 37: Need this for the `adsp` global variable to be set.
```csharp
lua.RegisterFunctions(...)
```
Let's create a stub class with no functions to start to use instead of `ScriptFunctions`.

solittlework: Hi @tzuge, I added lua.RegisterFunctions(null) in the TestScript function.

apps/script-service/Controllers/RunScriptRequest.cs
solittlework: You are right. I've deleted it. Thanks!

apps/script-service/Services/IScriptFunctions.cs
```diff
@@ -0,0 +1,18 @@
+
+using Adsp.Platform.ScriptService.Services.Platform;
+using Adsp.Platform.ScriptService.Services.Util;
+using Adsp.Sdk;
+using NLua;
```
tzuge, line 5: I think there are some unused imported namespaces here

apps/script-service/Services/StubScriptFunctions.cs
```diff
@@ -0,0 +1,33 @@
+using Adsp.Platform.ScriptService.Services.Platform;
+using Adsp.Platform.ScriptService.Services.Util;
+using Adsp.Sdk;
+using NLua;
+using RestSharp;
```
tzuge, line 5: guessing this is unused?

apps/script-service/Services/IScriptFunctions.cs
solittlework: Hi @tzuge, the unused namespaces have been removed. Thank you!

tzuge (approved):

PR #1352, opened 2022-11-14
Update the email template padding

solittlework:


libs/notification-shared/src/templates/email-wrapper.hbs
```diff
@@ -172,6 +172,8 @@
       .email-content {
         padding-left: 2.5rem;
         padding-right: 2.5rem;
+        padding-top: 2.5rem;
+        padding-bottom:2.5rem;
```
athena-chen-chen, line 176: Could be simplified with padding: 2.5rem 2.5rem 2.5rem 2.5rem;

athena-chen-chen (approved):

PR #1359, opened 2022-11-15
fix the access service page refresh issue

solittlework:


athena-chen-chen (approved):

apps/tenant-management-webapp/src/app/pages/admin/services/access/overview.tsx
```diff
@@ -36,15 +36,15 @@ export const Overview = (): JSX.Element => {
...
     return function clean() {
       if (isReady(indicator, userCount)) {
         dispatch(accessReset());
       }
     };
-  }, [indicator]);
+  }, []);
+
+  // eslint-disable-next-line
+  useEffect(() => {}, [indicator]);
```
smettu007, line 47: do we need this?

solittlework: Hi @smettu007, thank you for your review. I think we need this - we need to refresh the page when the value is updated. It determines the ready function value, which further determines page clean-up.

smettu007 (approved):

PR #1425, opened 2022-12-15
fix:Template editing code completion should only trigger for . when w…

athena-chen-chen:
…ithin handlebars expression

### DIFF
--- a/apps/tenant-management-webapp/src/app/lib/autoComplete.ts
+++ b/apps/tenant-management-webapp/src/app/lib/autoComplete.ts
@@ -1,11 +1,13 @@
 import type { EventDefinition } from '@store/event/models';
 import { languages, editor, Position } from 'monaco-editor';
 import { Monaco } from '@monaco-editor/react';
+
 export interface EditorSuggestion {
   label: string;
   insertText: string;
   children?: EditorSuggestion[];
 }
+const MAX_LINE_LENGTH = 10000;
 
 const createSuggestion = (instance: Monaco, suggestion: EditorSuggestion, hasBracket: boolean, hasParent = false) => {
   const showSuggest = suggestion.children?.length;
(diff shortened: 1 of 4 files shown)


apps/tenant-management-webapp/src/app/lib/autoComplete.ts
```diff
@@ -198,3 +199,21 @@ const getElementSuggestion = (element) => {
...
   return suggest;
 };
+const getPositionX = (text: string, search: string): number => {
+  const textArr = text.split('\n');
+  for (let i = textArr.length - 1; i >= 0; i--) {
+    if (textArr[i].indexOf(search) > -1) {
+      return i + 1;
+    }
+  }
+  return 10000;
```
rstyan, line 209: Could you make this a constant and give it an appropriate name?  Something that would tell us what 10000 means, like MAX_LINE_LENGTH.

apps/tenant-management-webapp/src/app/lib/autoComplete.ts
```diff
@@ -198,3 +199,21 @@ const getElementSuggestion = (element) => {
...
+};
+export const triggerInScope = (text: string, position: Position) => {
+  if (text.indexOf('{{') !== -1) {
+    if (getPositionX(text, '{{') <= position.lineNumber && getPositionX(text, '{{') >= position.lineNumber) {
+      return true;
+    }
+    return false;
+  }
+  return false;
+};
```
rstyan, line 219: This would be more succinct and easier to read if you said
return text.indexOf('{{') !== -1 &&
           getPositionX(text, '{{') <= position.lineNumber &&
           getPositionX(...) >= position.lineNumber
           
 In fact, having written it out, did you mean petPosition(text, '}}') >= position.lineNumber?
 Could you write a unit test for this method?

rstyan (changes requested):

apps/tenant-management-webapp/src/app/lib/autoComplete.ts
athena-chen-chen: set a MAX_LINE_LENGTH

apps/tenant-management-webapp/src/app/lib/autoComplete.ts
athena-chen-chen: New code be committed, please give another review, Thanks

rstyan (approved):

PR #1499, opened 2023-02-15
Added new actions in subscriptions redux to toggle show/hide subscriptions

smettu007:
[demo.webm](https://user-images.githubusercontent.com/31360331/219108729-79e94a31-9141-439f-8c87-4f7ef6bf3186.webm)


solittlework (approved):
The PR is generally good. But, we need to estimate whether it is good to include the UI state within a data object.

PR #1509, opened 2023-02-20
CS-1956-1976

jonathanweyermann:
image

### DIFF
--- a/apps/pdf-service/src/puppeteer.spec.ts
+++ b/apps/pdf-service/src/puppeteer.spec.ts
@@ -52,12 +52,12 @@ describe('puppeteer', () => {
               header
             </p>
           </div>`,
-        css: `<style>
+        additionalStyles: `<style>
           div {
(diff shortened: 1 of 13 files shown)


apps/pdf-service/src/puppeteer.spec.ts
```diff
@@ -52,6 +52,12 @@ describe('puppeteer', () => {
               header
             </p>
           </div>`,
+        additionalStyles: `<style>
+          div {
+            font-size: 32px;
```
smettu007, line 57: is this the default size that we are using across the pdf?

jonathanweyermann: No, it's just some arbitrary size for testing purposes

smettu007: oh!  my bad! its a spec file

smettu007 (approved):

PR #1510, opened 2023-02-20
subscriber-app: fixed 404 error check

smettu007:
Screenshot 2023-02-20 at 3 03 41 PM


jonathanweyermann (approved):
What if there is some sort of unknown 400 error

PR #1529, opened 2023-03-03
tenant-web: subscribers list code cleanup

smettu007:


solittlework (approved):
Thanks, Subbu. This PR improves the code structure. There are a few things we need to address in the feature. 
1) The openModalFunction and openDeleteModalFunction will cause unnecessary renders since it changes their parent state. It will become worse if we include dispatch here.
2) Passing a Subscriber (an object) is not a good idea. Mostly, when updating the reducer, we create a new object.
The comments above are not related to this PR. But, it might relate to the task.

smettu007:
> Thanks, Subbu. This PR improves the code structure. There are a few things we need to address in the feature.
> 
> 1. The openModalFunction and openDeleteModalFunction will cause unnecessary renders since it changes their parent state. It will become worse if we include dispatch here.
> 2. Passing a Subscriber (an object) is not a good idea. Mostly, when updating the reducer, we create a new object.
>    The comments above are not related to this PR. But, it might relate to the task.

The idea is, with this story https://goa-dio.atlassian.net/browse/CS-1989 we wont be needing dispatches anymore in a child component.
As for opening and closing modal, i will review them and see how it impacts the parent container. 
We dont need to pass the whole subscriber object. I will clean it up as part of my investigation.
This PR is simply to re-structure the code for better readability

PR #1550, opened 2023-03-07
fix: pdf add edit modal UI improvement and e2e in pdf and file service

athena-chen-chen:
Screen Shot 2023-03-07 at 1 31 49 PM


solittlework (approved):
The PR looks good, except for a small lint issue for the lodash.

PR #1573, opened 2023-03-13
CS-2049

jonathanweyermann:
separate pdf generation components from editor and reduce chances of random additional propagation propagation

### DIFF
--- a/apps/tenant-management-webapp/src/app/pages/admin/services/pdf/templates/previewEditor/PreviewTemplate.tsx
+++ b/apps/tenant-management-webapp/src/app/pages/admin/services/pdf/templates/previewEditor/PreviewTemplate.tsx
@@ -1,5 +1,6 @@
 import React, { FunctionComponent, useState, useEffect } from 'react';
 import { GoAButton, GoAIconButton } from '@abgov/react-components-new';
+import _ from 'underscore';
 
 import {
   SpinnerPadding,
@@ -12,7 +13,9 @@ import {
 import { RootState } from '@store/index';
 import { useSelector, useDispatch } from 'react-redux';
 import { DownloadFileService } from '@store/file/actions';
+import { streamPdfSocket } from '@store/pdf/action';
 import { GoAPageLoader } from '@abgov/react-components';
+
 interface PreviewTemplateProps {
   channelTitle: string;
   generateTemplate: () => void;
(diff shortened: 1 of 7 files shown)


apps/tenant-management-webapp/src/app/pages/admin/services/pdf/templates/previewEditor/PreviewTemplate.tsx
```diff
@@ -43,17 +46,22 @@ export const PreviewTemplate: FunctionComponent<PreviewTemplateProps> = ({ chann
...
 
-  const indicator = useSelector((state: RootState) => {
-    return state?.session?.indicator;
-  });
+  const indicator = useSelector((state: RootState) => state?.session?.indicator, _.isEqual);
 
-  const files = useSelector((state: RootState) => {
-    return state?.pdf.files;
-  });
+  const files = useSelector((state: RootState) => state?.pdf.files, _.isEqual);
```
solittlework, line 51: If I would use the 'reselect' rather than introduce the _.isEqual.

apps/tenant-management-webapp/src/app/pages/admin/services/pdf/templates/previewEditor/PreviewTemplate.tsx
```diff
@@ -43,17 +46,22 @@ export const PreviewTemplate: FunctionComponent<PreviewTemplateProps> = ({ chann
 
   const dispatch = useDispatch();
 
-  const indicator = useSelector((state: RootState) => {
-    return state?.session?.indicator;
-  });
+  const indicator = useSelector((state: RootState) => state?.session?.indicator, _.isEqual);
```
solittlework, line 49: It is not a good idea to use indicators at this level. For example, if we like to make the button in the template disabled when an action takes place. The best way in my mind is to move the button to a separate component. We can add 'indicator' in that separate component to avoid unnecessary refresh. Another very similar issue is that {indicaot.show && <Any loading indicator> } is not suggested. To avoid the issue, the global page indicator in the components/indicator.tsc comes with an useSelect function to detect the state update. I would wish the action loading indicator in the Page has a similar structure here. I think using _isEqual is not unnecessary here.

solittlework (approved):

PR #1654, opened 2023-04-28
CS 2038 - 

jonathanweyermann:
File service in the job mode runs scheduled jobs

The retention policy file deletion is a scheduled job

- runs on daily
- operation performed by the job
- - Load all tenants
- - For each tenant, load file configuration (and core file configuration)
- - For each file type with retention policy enabled, find files that should be deleted
- - Delete files

* Job handles paging files when querying

Deletion functions as regular deletion and associated domain events are emitted where ‘user’ information indicates the retention process

mark for deletion (doesn’t need to be immediately deleted from underlying storage)

### DIFF
--- a/apps/file-service/src/file/events.ts
+++ b/apps/file-service/src/file/events.ts
@@ -108,6 +108,7 @@ export function fileDeleted(user: User, file: File): DomainEvent {
       deletedBy: {
         id: user.id,
         name: user.name,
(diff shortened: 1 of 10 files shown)


athena-chen-chen:
Hi, Jonathan, could you remove comments and fix lint warning, Thanks!

PR #1656, opened 2023-04-29
CS-2061

jonathanweyermann:
* Reset button wasn't working anyone due to using different dropdown ui component - Now using react to reload the component on reset in order to set it back to the initial state
* Made both file upload components the same - placed text and button side by side


rstyan (approved):
LGTM

PR #1660, opened 2023-05-02
CS-2038 - Events show retention days

jonathanweyermann:
* Events now show retention days correctly
* Correctly counting number of events now in log


rstyan (approved):
LGTM

PR #1778, opened 2023-06-21
CS-2088 - WaitTime now is saved in status configuration

jonathanweyermann:
* Some ui fixes in test window
* Filled in some missing information in webhook event


rstyan (approved):
LGTM

PR #1779, opened 2023-06-22
Webhook fixes

jonathanweyermann:
* Use expected icon
* hide test data in edit page
* Fix error that occurred when there were no webhooks / show no webhook message
image


rstyan (approved):
LGTM

PR #1791, opened 2023-07-01
CS-2080

jonathanweyermann:
used SDK to grab webhooks, had to store them one level up

### DIFF
--- a/apps/push-service/src/push/configuration.ts
+++ b/apps/push-service/src/push/configuration.ts
@@ -34,9 +34,9 @@ export const configurationSchema = {
             required: ['namespace', 'name'],
           },
         },
-        webhooks: {
-          type: 'array',
-          items: {
+        webhook: {
+          type: 'object',
+          properties: {
             id: { type: 'string' },
             name: { type: 'string' },
             url: { type: 'string' },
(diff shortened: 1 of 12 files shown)


apps/push-service/src/push/configuration.ts
```diff
@@ -34,27 +53,10 @@ export const configurationSchema = {
...
-              },
-            },
-          },
-        },
         subscriberRoles: { type: 'array', items: { type: 'string' } },
         publicSubscribe: { type: 'boolean' },
       },
-      required: ['id', 'name'],
+    
     },
```
solittlework, line 60: It looks we change the webhooks type. Do we need to delete the old data before deployment?

apps/push-service/src/push/router/stream.ts
```diff
@@ -255,94 +266,86 @@ export const createStreamRouter = (
...
+          let callResponseTime = 0;
+          const beforeWebhook = new Date().getTime();
+          try {
+            if (isValidUrl(endpointWebsocket)) {
+              response = await axios.post(endpointWebsocket, next);
+              callResponseTime = new Date().getTime() - beforeWebhook;
+            }
+          } catch (err) {
+            response.statusText = err.message;
+            response.status = 400;
```
solittlework, line 322: Ideally, the error response status shall be based on the error type. We can think of how to improve this laster.

solittlework (approved):
The PR wrappers the webhook into a specific property. Except for the lints warnings, the PR is generally good.

PR #1843, opened 2023-07-21
Add more try catch to the status service

solittlework:


solittlework:
We just added more logs to debug the status service issues on UAT and I am going to merge it directly.

PR #1860, opened 2023-07-31
Update the file service job log

solittlework:


solittlework:
This PR only enriches the log as requested. I am going to merge it.

PR #1864, opened 2023-08-01
Fix the CodeQL syntax error

solittlework:


solittlework:
Fix the CodeQL syntax error

PR #1877, opened 2023-08-08
Fix:revision-add-icon

RamaChandiniDevarapalli:


apps/tenant-management-webapp/src/app/pages/admin/services/configuration/revisions/revisionsTable.tsx
```diff
@@ -165,6 +165,13 @@ const RevisionTableComponent: FunctionComponent<RevisionTableComponentProps> = (
       dispatch(getConfigurationActive(service));
     }
   }, [indicator, revisions]);
+  if (
+    !configurationRevisions[service]?.revisions.next &&
+    !configurationRevisions[service]?.revisions?.latest &&
+    configurationRevisions[service]?.revisions.result.length
```
athena-chen-chen, line 171: It better configurationRevisions[service]?.revisions?.result?.length to avoid future bug.

RamaChandiniDevarapalli:
> Remove the unused variable 'activeEditState' if it is not needed.

@thdlam I think this warning is not from this component from forms i think

apps/tenant-management-webapp/src/app/pages/admin/services/configuration/revisions/revisionsTable.tsx
RamaChandiniDevarapalli: updated change

PR #1884, opened 2023-08-08
Roll back the pdf service dockerfile

solittlework:


solittlework:
Revert the pervious update

PR #1924, opened 2023-08-14
Cs 2170

jonathanweyermann:
image


apps/tenant-management-webapp/src/app/pages/admin/services/form/definitions/addEditFormDefinitionEditor.tsx
```diff
@@ -0,0 +1,368 @@
...
+    }
+
+    window.addEventListener('resize', handleResize);
+    return () => window.removeEventListener('resize', handleResize);
+  }, []);
+
+  return windowDimensions;
+}
+
+export const AddEditFormDefinitionEditor: FunctionComponent = () => {
```
solittlework, line 65: I would suggest using the JSX.Element here. We shall try to deprecate unnecessary functionComponent.

solittlework (approved):

PR #1952, opened 2023-08-21
CS-2170 - Full screen editor

jonathanweyermann:
image

* Full screen form
* Name/Definition/Description goes into separate edit modal
* Editor is on the left / Form Permissions are on the right
* creating a new one shows Name/Definition/Description modal first, and then goes to full screen modal with the Schema and Form Permissions
* hitting back after making changes shows a model that asks you to save or cancel


solittlework (approved):

solittlework (changes requested):
The code comes with a build error.

jonathanweyermann:
Should be good now

solittlework (approved):

PR #1994, opened 2023-08-28
Fix :attribute id's-fix

RamaChandiniDevarapalli:


apps/tenant-management-webapp/src/app/components/AsideLinks.tsx
```diff
@@ -14,12 +14,12 @@ function AsideLinks({ serviceLink, docsLink }: AsideLinksProps): JSX.Element {
       <GapAdjustment>Helpful links</GapAdjustment>
       {serviceLink && (
         <HyperLinkColor>
-          <ExternalLink link={serviceLink} text="See the code" />
+          <ExternalLink testId="code-link" link={serviceLink} text="See the code" />
```
thdlam, line 17: Should this be data-testid?

apps/tenant-management-webapp/src/app/components/AsideLinks.tsx
```diff
@@ -14,12 +14,12 @@ function AsideLinks({ serviceLink, docsLink }: AsideLinksProps): JSX.Element {
...
         </HyperLinkColor>
       )}
       {docsLink && (
         <HyperLinkColor>
-          <ExternalLink link={docsLink} text="Read the API docs" />
+          <ExternalLink link={docsLink} testId="docs-link" text="Read the API docs" />
```
thdlam, line 22: Should this be data-testid

apps/tenant-management-webapp/src/app/components/AsideLinks.tsx
RamaChandiniDevarapalli: The property added to the external link is testId

thdlam (approved):

apps/tenant-management-webapp/src/app/components/AsideLinks.tsx
solittlework: I just checked the code and it looks the ExternalLink has the "testId" property.

apps/tenant-management-webapp/src/app/components/AsideLinks.tsx
RamaChandiniDevarapalli: interface ExternalLinkProps {
  text: string;
  link: string;
  testId?: string;
}
testId is the property in ExternalLink

PR #2005, opened 2023-08-30
Fix : review fixes

RamaChandiniDevarapalli:


apps/tenant-management-webapp/src/app/pages/admin/services/pdf/styled-components.ts
```diff
@@ -40,7 +40,7 @@ export const BadgeWrapper = styled.div`
 `;
 
 export const ButtonBox = styled.div`
-  width: 2.15rem;
+  width: 1.75rem;
```
solittlework, line 43: I would suggest mapping the width to the most closed css variable.

solittlework (approved):

apps/tenant-management-webapp/src/app/pages/admin/services/pdf/styled-components.ts
RamaChandiniDevarapalli: I thought to but we are missing exact alignment with the other controls. So I made change without css variable

PR #2022, opened 2023-09-05
Fix the errors in the stream service

solittlework:


apps/tenant-management-webapp/src/app/components/RoleTable.tsx
```diff
@@ -48,8 +49,9 @@ export const ClientRoleTable = (props: ClientRoleTableProps): JSX.Element => {
   };
 
   const getClientId = () => {
-    return props.clientId ? <PaddingRem>{props.clientId}</PaddingRem> : tenantName;
+    return props.clientId && props.clientId !== REALM_ROLE_KEY ? props.clientId : tenantName;
```
jonathanweyermann, line 52: can you put <PaddingRem> around props.clientId

apps/tenant-management-webapp/src/app/components/styled-components.ts
```diff
@@ -2,11 +2,7 @@ import styled from 'styled-components';
 
 export const MarginAdjustment = styled.h4`
   margin-bottom: 0.5rem !important;
-  margin-top: 2rem !important;
-`;
-
-export const PaddingRem = styled.div`
-  padding-top: 1rem;
+  margin-top: 3rem !important;
 `;
```
jonathanweyermann, line 6: don't do this change

jonathanweyermann (changes requested):
couple comments

apps/tenant-management-webapp/src/app/components/styled-components.ts
solittlework: Sure, I will roll back it.

jonathanweyermann (approved):

PR #2043, opened 2023-09-12
needs to be an exact match

jonathanweyermann:


apps/tenant-management-webapp/src/app/pages/admin/services/notifications/subscription/subscriptionList.tsx
```diff
@@ -197,15 +197,11 @@ const SubscriptionsListComponent: FunctionComponent<SubscriptionsListComponentPr
...
               <tr>
-                <th
-                  className="address-as"
-                  id={`userName_${index}`}
-                  data-testid={`subscription-header-address-as-${index}`}
-                >
+                <th className="address-as" id="userName" data-testid={`subscription-header-address-as-${index}`}>
                   Address as
                 </th>
-                <th id={`channels_${index}`}>Channels</th>
```
thdlam, line 207: Hi Jonathan, can you run the axe DevTools and make sure there are no  errors for the header id field, i think we might still need a unique key for the id attribute.

jonathanweyermann: hmm. I missed one

jonathanweyermann: There are no errors, but I'd like to verify in uat

thdlam (approved):

PR #2050, opened 2023-09-13
Fix : delaying queue to configure

RamaChandiniDevarapalli:


jonathanweyermann (commented):
Can you explain your reasoning this is required? Generally we'd probably like to avoid artificial delays unless we can't help it.

RamaChandiniDevarapalli:
closing this PR as there is any alterbative way to solve thus

PR #2051, opened 2023-09-13
CS-2393/ CS-2394 / CS-2395

jonathanweyermann:
image
image
image

### DIFF
--- a/apps/tenant-management-webapp/src/app/pages/admin/services/comment/topicTypes/topicTypes.tsx
+++ b/apps/tenant-management-webapp/src/app/pages/admin/services/comment/topicTypes/topicTypes.tsx
@@ -1,0 +1,115 @@
+import React, { useEffect, useState } from 'react';
+
+import { GoAButton } from '@abgov/react-components-new';
+
+import { useDispatch, useSelector } from 'react-redux';
+import { getCommentTopicTypes, updateCommentTopicType } from '@store/comment/action';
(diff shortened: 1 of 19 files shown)


solittlework (approved):
Hi @jonathanweyermann, except for a few lint issues, the rest are good.

apps/tenant-management-webapp/src/app/pages/admin/services/comment/topicTypes/topicTypes.tsx
```diff
@@ -0,0 +1,115 @@
...
+  }, []);
+
+  const reset = () => {
+    setOpenAddCommentTopicType(false);
+  };
+
+  // eslint-disable-next-line
+  useEffect(() => {}, [commentDefinitions]);
+  useEffect(() => {
+    document.body.style.overflow = 'unset';
```
thdlam, line 59: I think we should create common function for this, even though it is a one liner to show the scrollbar. 
It is a couple of files now.

thdlam (approved):

apps/tenant-management-webapp/src/app/pages/admin/services/comment/topicTypes/topicTypes.tsx
thdlam: ok, its fine then. Everything looks good.

PR #2068, opened 2023-09-15
CS-2444 - Hide scrollbars in modals and hide space between banner and modal

jonathanweyermann:


solittlework (commented):
I have concerns about adding global overflow style in this way. Ideally, we shall allow different overflow in different cases or pages. I would believe GoAModal is able to clean up the overflow styles by itself as we did in the event service.

jonathanweyermann:
Where in event service are we cleaning up overflow styles on a full screen? I'm trying to hide the scrollbar and make it appear again on a full screen modal. The full screen modals have always been set to hide the scrollbar because it's weird to display it, but we didn't always have a clean way to unhide. 

I could set it back to

` useEffect(() => {
    document.body.style.overflow = 'unset';
  }, []);`

but I do need to have something there to not ruin the scrollbar after a modal

jonathanweyermann:
Going to try without manually adjusting any overflow - we may be overthinking this

PR #2095, opened 2023-09-22
Prevent reloading of data from saga

thdlam:


solittlework (commented):
Hi @thdlam , I suggest double check wether we can avoid the loadash here. lodash is a heavy lib: https://youmightnotneed.com/lodash. Please let me know if you really need it.

thdlam:
> Hi @thdlam , I suggest double check wether we can avoid the loadash here. lodash is a heavy lib: https://youmightnotneed.com/lodash. Please let me know if you really need it.

I will write a simple function to check if an object is empty. Didnt realize this is an issue with lodash.

solittlework (approved):

PR #2144, opened 2023-10-05
Feat/file secure classifications

thdlam:
Security Data classification for file types
image


apps/file-service/src/file/configuration/schema.ts
```diff
@@ -9,6 +9,11 @@ export const configurationSchema = {
         anonymousRead: { type: 'boolean' },
         readRoles: { type: 'array', items: { type: 'string' } },
         updateRoles: { type: 'array', items: { type: 'string' } },
+        securityClassification: {
+          type: 'string',
+          enum: ['public', 'protected a', 'protected b', 'protected c'],
+          default: 'protected a',
```
solittlework, line 15: I would suggest double checking the default value here. I am not sure whether we need to track the old file without setting the file security. If we not care about it, I would say using protected a here is correct.

solittlework (approved):

apps/file-service/src/file/configuration/schema.ts
thdlam: The default value will be protected a in case they dont put anything in security classification (if they make a web api to this endpoint to update the schema this will enforce the data structure)
Old file types wont be enforce until they update or create a new file type.

However, from the web app point of view, default is protected a when they create or update the file type.

PR #2161, opened 2023-10-10
Fix : Delete queue

RamaChandiniDevarapalli:


jonathanweyermann (changes requested):
Before you merge this, can you confirm that closing without deleting does not remove the scrollbar - I've noticed when delete is moved into a sub-component it tends to cause issues with the scrollbar

jonathanweyermann (approved):

apps/tenant-management-webapp/src/app/pages/admin/services/task/deleteConfirmationsView.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+import { TableDiv } from './styled-components';
+import { DeleteModal } from '@components/DeleteModal';
+
+import { deleteTaskQueue } from '@store/task/action';
+
+import { RootState } from '@store/index';
+
+interface calendarTableProps {
+  queue;
```
thdlam, line 14: is data type you can add onto this?

apps/tenant-management-webapp/src/app/pages/admin/services/task/queuesList.tsx
```diff
@@ -90,44 +84,27 @@ export const QueuesList = ({ openAddDefinition }: AddEditQueueProps): JSX.Elemen
...
-      {!indicator.show && Object.keys(taskQueues).length > 0 && (
+      {Object.keys(taskQueues).length > 0 && (
         <QueueListTable
           taskQueues={taskQueues}
           onDelete={(currentTemplate) => {
-            setShowDeleteConfirmation(true);
             setSelectedQueue(currentTemplate);
+
+            setTimeout(() => {
+              setShowDeleteConfirmation(true);
```
thdlam, line 100: is there another approach you can do with this instead of using the setTimeout?

PR #2192, opened 2023-10-19
Feat/cs 2449 adding security classification to topics

RamaChandiniDevarapalli:


apps/comment-service/migrations/20230823152146_create_comment_tables.js
solittlework, line 1: Hi @RamaChandiniDevarapalli for the db schema migration, we are expected to change the previous history, which will cause the migration conflicts.  Please create a new migration script and add the new column in the script.

solittlework (changes requested):

apps/comment-service/migrations/20230823152146_create_comment_tables.js
RamaChandiniDevarapalli: Hi I didn't change any schema in previous migration I just added a condition to check exists or not

thdlam: Hi @RamaChandiniDevarapalli I dont think you need to create a column in the topics table. I think you are suppose to get the securityClassifcation from the Topic Types table and return the topic with the security classification.

If you add the securityClassification to the topics table and then when some one changes the securityClassification on the file type, The topics and and topic type will incorrect security classification.

@jonathanweyermann is what I am saying correct for securityClassification or am i out to lunch on this?

solittlework (approved):

PR #2200, opened 2023-10-21
Cs 2471 - Detect file mime type on upload and persist result as part of the record

jonathanweyermann:
- File mime type is detected when a file is uploaded
- Mime type is stored as part of the file record
- Mime type is returned as part of the file record (metadata)
- When file is downloaded, mime type is set on appropriate headers
- For backwards compatibility, defaults to application/octet-stream when mime type not in record
- If detection fails, then default to not value in the record, and send download response as application/octet-stream
- file upload streams the content into a backend store, and mime type detection should have minimal impact on this process. e.g. in terms of back pressure, etc.


tzuge:
I would not expect to see the mime type detection in azure-blob.ts. The mime type detection process should be agnostic of the specific storage provider. If that means the Readable interface on entity.content is no longer sufficient, then we can sub-class it for the extra mime type info.

solittlework (approved):

thdlam (approved):

PR #2249, opened 2023-11-06
Feat/cs 2341 design change to script

RamaChandiniDevarapalli:


apps/tenant-management-webapp/src/app/pages/admin/services/script/editor/tombstone.tsx
```diff
@@ -0,0 +1,126 @@
...
+import { AddScriptModal } from '../addScriptModal';
+
+import { useSelector } from 'react-redux';
+import { tenantRolesAndClients } from '@store/sharedSelectors/roles';
+
+interface props {
+  selectedScript: ScriptItem;
+  onSave;
+}
+export const TombStone: FunctionComponent<props> = ({ selectedScript, onSave }) => {
```
solittlework, line 15: A very small issue here. I have not see children property here. So, I would suggest replacing the FunctionComponent with normal Element.JSX.

solittlework (approved):

apps/tenant-management-webapp/src/app/pages/admin/services/script/editor/scriptEditor.tsx
```diff
@@ -65,6 +81,10 @@ export const ScriptEditor: FunctionComponent<ScriptEditorProps> = ({
   const monaco = useMonaco();
   let activeParam = 0;
   let activeSignature = 0;
+
+  useEffect(() => {
+    setScript(selectedScript);
+  }, [selectedScript]);
```
solittlework, line 87: The selectedScript is a property. I am not sure what is the reason of detecting a selectedScript here.

PR #2251, opened 2023-11-07
CS-2582 - SMS And Phone validation

jonathanweyermann:
image
image
image
image


apps/notification-service/src/verify.ts
```diff
@@ -44,28 +49,39 @@ export class VerifyServiceImpl implements VerifyService {
...
 
+    if (!expireIn) {
+      expireIn = 10;
+    }
+
     const verifyServiceUrl = await this.directory.getServiceUrl(adspId`urn:ads:platform:verify-service`);
 
     const key = uuidv4();
-    const verifyRequestUrl = new URL(`/verify/v1/codes/${key}`, verifyServiceUrl);
+    const verifyRequestUrl = new URL(`/verify/v1/codes/${key}?expireIn=${expireIn}`, verifyServiceUrl);
```
solittlework, line 59: Do we need to show the expireIn in the api? Will it actually control the exp time? I think we need to include the redirect URL here.

jonathanweyermann: the way the verify api is coded, the expireIn needs to be a param. maybe I should change the verify api to use the expireIn in the body instead ?

jonathanweyermann: Is it works correctly this way without changing verify api, but maybe verify api is coded badly

PR #2278, opened 2023-11-10
CS-2544 - Verify Channel Modal

jonathanweyermann:
* Clean up ui and create a separate verify channel modal
* Fix phone flow
image
image

should fix CS-2613 as well


solittlework (approved):
I checked the logic part of the PR and it looks good. But, there are a few lint issues in the PR. I hope we can fix them before merge.

PR #2280, opened 2023-11-11
CS-2545

jonathanweyermann:
image


athena-chen-chen (approved):

thdlam (approved):

apps/subscriber-app/src/app/notificationBanner.tsx
```diff
@@ -38,4 +54,14 @@ export const NotificationStyles = styled.div`
     margin: 0 !important;
     padding: 0 !important;
   }
+
+  display: flex;
+
+  .verification-code-button {
+    margin-left: 8px;
+    margin-top: -4px;
```
solittlework, line 62: Hi @jonathanweyermann, is there any reason we try to avoid using the styled-component here? I remember the GoA CSS variables have been input to the app. I would suggest double checking it.

apps/subscriber-app/src/app/app.tsx
```diff
@@ -17,7 +17,7 @@ const AppRouters = () => {
   return (
     <Router>
       <Routes>
-        <Route path="/subscriptions/:realm" element={<PrivateApp />} />
+        <Route path="/subscriptions/:realm/*" element={<PrivateApp />} />
```
solittlework, line 20: I'd like to confirm one thing is that are we going to fix some previous routing issue here? I am asking this question because when we add a wild card, I expect we will introduce some new routes within the privateApp. But, I have not seen any update there.

jonathanweyermann: Sorry, this is a leftover from a previous PR - Currently there is a flow that uses the phone to verify code that is currently broken in dev/uat because I'm burying the route by not doing this

apps/subscriber-app/src/app/notificationBanner.tsx
jonathanweyermann: I don't think I can get 4px using CSS variables, but If I can I'd like to fix it in a follow up PR - this pr contains from preliminary styling that I probably have to fix later anyway

solittlework (approved):

PR #2295, opened 2023-11-17
Update API docs app to use a scheduled job to update aggregate API docs once per day

athena-chen-chen:


apps/api-docs-app/src/docs/serviceDocs.ts
```diff
@@ -111,7 +109,8 @@ class ServiceDocsImpl {
     const namespaces = this.cache.keys();
     let mergedDocs: Record<string, ServiceDoc> = {};
     const docs: Record<string, ServiceDoc> = {};
-
+    console.log('id', id);
```
thdlam, line 112: do we need these console.logs?

thdlam (approved):

apps/api-docs-app/src/docs/serviceDocs.ts
thdlam: looks good except for the console logs that i am wondering about.

athena-chen-chen: Thanks, I remove those console.log.

PR #2383, opened 2023-12-12
Feat:Delete topic and comment functionality

RamaChandiniDevarapalli:


apps/tenant-management-webapp/src/app/pages/admin/services/comment/topics/deleteConfirmationsView.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { TopicItem } from '@store/comment/model';
+
+interface calendarTableProps {
+  topic: TopicItem;
+  selectedType: string;
+  onCancel?: () => void;
+  onDelete?: (topic) => void;
+}
+
+export const DeleteConfirmationsView: FunctionComponent<calendarTableProps> = ({
```
solittlework, line 18: Hi @RamaChandiniDevarapalli , I think we can use the better way to pass the properties here. Otherwise, I feel we will do another around code refactoring in the near future. And, the FunctionComponent is something we need to avoid.

apps/tenant-management-webapp/src/app/store/comment/action.ts
```diff
@@ -116,7 +116,7 @@ export interface AddCommentAction {
 
 export interface DeleteCommentAction {
   type: typeof DELETE_COMMENT;
-  payload: string;
+  payload: object;
```
solittlework, line 119: Maybe, DeletePayload here?

solittlework (approved):

PR #2388, opened 2023-12-13
Fix: bug fixes in comment and task

RamaChandiniDevarapalli:


apps/comment-service/src/postgres/topic.ts
```diff
@@ -77,7 +77,7 @@ export class PostgresTopicRepository implements TopicRepository {
       results: rows.map((r) => this.mapRecord(types, r)),
       page: {
         after,
-        next: encodeNext(rows.length, top, skip),
+        next: encodeNext(rows.length, topChecked, skip - 1),
         size: rows.length,
```
solittlework, line 81: I think it is the "size" rather than the topChecked part has issue. How about changing the size to 

        size: rows.length > top ? top : rows.length,

RamaChandini: updated change.

solittlework (approved):

PR #2390, opened 2023-12-14
Cs 2652 - Multi-step form

jonathanweyermann:
image


apps/tenant-management-webapp/src/app/pages/admin/services/form/definitions/formStepper.tsx
```diff
@@ -0,0 +1,160 @@
...
+  const [step, setStep] = useState<number>(-1);
+  const [stepData, setStepData] = useState(data);
+
+  function setPage(page) {
+    if (page < 1 || page > uiSchema.elements?.length + 1) return;
+    setStep(page);
+  }
+
+  const renderers = [
+    ...materialRenderers,
```
rstyan, line 30: You're not really making use of materialRenderers here; I mean, couldn't you just as well use vanillaRenderers?  I don't really know if it makes much difference except that material-ui is a large package and if it is unnecessary why include it?

apps/tenant-management-webapp/src/app/pages/admin/services/form/definitions/formStepper.tsx
```diff
@@ -0,0 +1,160 @@
...
+    {
+      tester: categorizationRendererTester,
+      renderer: FormStepperControl,
+    },
+  ];
+
+  return (
+    <div id="#/properties/formStepper" className="formStepper">
+      <GoAFormStepper testId="form-stepper-test" step={step} onChange={(step) => setStep(step)}>
```
rstyan, line 39: From a clean-code perspective, it would be nice to see these elements in separate files:
GoAFormStepper, GoAPages, ReviewItem, etc.

apps/tenant-management-webapp/src/app/pages/admin/services/form/definitions/Rating.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { useState } from 'react';
+import { InputLabel } from '@mui/material';
+
+interface RatingProps {
+  id?: string;
+  value: number;
+  updateValue: (newValue: number) => void;
+}
+
+export const Rating: React.FC<RatingProps> = ({ id, value, updateValue }) => {
```
rstyan, line 11: This isn't for our production version, right?

rstyan (dismissed):
Overall it looks great.  I added a couple of comments for the non-spike version that would get merged.

apps/tenant-management-webapp/src/app/pages/admin/services/form/definitions/formStepper.tsx
jonathanweyermann: I think it's for elements that we don't have a wrapper for, but I could try without it

apps/tenant-management-webapp/src/app/pages/admin/services/form/definitions/Rating.tsx
jonathanweyermann: Oh, yeah. Anything with the word rating in it is can be removed

jonathanweyermann:
I'm on the plane and won't be able to do code changes. If someone wants to modify this pr and merge it I won't be offended. Otherwise I'll fix and merge it Tuesday

apps/tenant-management-webapp/src/app/pages/admin/services/form/definitions/formStepper.tsx
rstyan: That's true, but I think vanillaRenderers will do the trick.

libs/jsonforms-components/project.json
```diff
@@ -12,26 +12,6 @@
...
-      "executor": "@nrwl/rollup:rollup",
-      "outputs": ["{options.outputPath}"],
-      "options": {
-        "outputPath": "dist/libs/jsonforms-components",
-        "tsConfig": "libs/jsonforms-components/tsconfig.lib.json",
-        "project": "libs/jsonforms-components/package.json",
-        "entryFile": "libs/jsonforms-components/src/index.ts",
-        "external": ["react/jsx-runtime"],
-        "rollupConfig": "@nrwl/react/plugins/bundle-rollup",
-        "compiler": "babel",
```
solittlework, line 25: Hi @jonathanweyermann, do you have any specific reason to remove the build here.

solittlework (approved):

libs/jsonforms-components/project.json
jonathanweyermann: yeah, it caused a linting error - https://stackoverflow.com/questions/74887933/buildable-libraries-cannot-import-or-export-from-non-buildable-libraries-eslint

PR #2431, opened 2024-01-03
Fix-alignment fixes

RamaChandiniDevarapalli:


thdlam (commented):
Can you look at the styled components and change it to use design tokens

thdlam (approved):

PR #2434, opened 2024-01-03
CS-2468 - Add form submission database

jonathanweyermann:
CS-2468 - Add form submission - 

Allow multiple forms to be created based on CS-2452 

fulfills CS-2468 requirement - 
`Form service form submit operation uses configuration to create submission records if appropriate.`

### DIFF
--- a/apps/form-service/src/form/router/form.ts
+++ b/apps/form-service/src/form/router/form.ts
@@ -201,26 +201,6 @@ export function accessForm(notificationService: NotificationService): RequestHan
     }
   };
 }
-export function getSubmissionRecords(notificationService: NotificationService): RequestHandler {
-  return async (req, res, next) => {
-    try {
-      const end = startBenchmark(req, 'operation-handler-time');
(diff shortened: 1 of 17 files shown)


apps/form-service/src/form/router/form.ts
```diff
@@ -377,6 +407,15 @@ export function createFormRouter({
     getForm(repository),
     accessForm(notificationService)
   );
+  router.get(
+    '/forms/:formId/submissions/:submissionId',
+    createValidationHandler(
+      param('formId').isUUID(),
+      query('code').optional().isString().isLength({ min: 1, max: 10 })
```
thdlam, line 414: should this be submissionId instead of code?

jonathanweyermann: Oh, I think those lines are me getting ahead of myself (different ticket) - I'm going to remove them

thdlam (approved):

PR #2517, opened 2024-01-29
Upgrade to route-dom-6

RamaChandiniDevarapalli:


apps/tenant-management-webapp/src/app/app.tsx
```diff
@@ -33,40 +33,23 @@ const AppRouters = () => {
...
-              <Route exact path="/logout-redirect">
-                <LogoutRedirect />
-              </Route>
-              <Route exact path="/tenant/creation">
-                <CreateTenant />
-              </Route>
-            </PublicApp>
-          </Switch>
-        </Router>
+          <Route path="/admin//*" element={<PrivateApp />}>
```
thdlam, line 39: is this path correct? There is an extra backslash in it?

solittlework: I tested it locally and with or without the /* does not affect the final results.

solittlework (approved):

PR #2602, opened 2024-02-21
Submit button enable based on validation

RamaChandiniDevarapalli:


libs/jsonforms-components/src/lib/Controls/FormStepper/FormStepperControl.tsx
```diff
@@ -70,10 +70,25 @@ export const FormStepper = ({
...
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const vslidateFormData = (formData: Array<UISchemaElement>) => {
+    const validate = ajv.compile(schema);
+    return validate(formData);
+  };
+
+  useEffect(() => {
+    const valid = vslidateFormData(data);
+    setIsFormValid(valid);
```
solittlework, line 90: Hi Rama, we can leave the form validation to the future work. We shall not use the ajv compile directly here. Do we have the errors attribute in the props. We can use whether the 'errors' is empty or not to determine the form validation. If this is not the case, we can leave it to the future work.

solittlework (approved):
@RamaChandiniDevarapalli, the PR is generally good. But, I think use the ajv directly is not correct. You can leave it here and explain it to QA.

libs/jsonforms-components/src/lib/Controls/FormStepper/FormStepperControl.tsx
RamaChandiniDevarapalli: Discussed with @solittlework merging this and trying other way to validate data

PR #2683, opened 2024-03-08
fix: queue task and disposition that is not saving in form editor

thdlam:


apps/tenant-management-webapp/src/app/pages/admin/services/form/definitions/addEditFormDefinitionEditor.tsx
```diff
@@ -847,33 +855,26 @@ export function AddEditFormDefinitionEditor(): JSX.Element {
...
-          dispositionStates.push(currentDispositions);
-          tempDefinition.dispositionStates = dispositionStates;
-          setDefinition(tempDefinition);
+          if (newDisposition) {
+            const dispositionStates = definition.dispositionStates || [];
+            const tempDefinition = { ...definition };
+            dispositionStates.push(currentDispositions);
+            tempDefinition.dispositionStates = dispositionStates;
+            setDefinition(tempDefinition);
+          } else {
```
rstyan, line 869: To make this code cleaner you should refactor and create a function, i.e.
const saveDisposition = (currentDispositions, etc...) => { onSave code goes here}
and then onSave=saveDisposition
As it is, you have no way to unit test this code.

rstyan (changes requested):

apps/tenant-management-webapp/src/app/pages/admin/services/form/definitions/addEditFormDefinitionEditor.tsx
thdlam: Hi @rstyan, can you take another look at it.

rstyan (approved):
Nice.  This is exactly what was needed.

PR #2689, opened 2024-03-11
feat: review page edit functionality

RamaChandiniDevarapalli:


solittlework (approved):

rstyan (changes requested):
Rama, we're trying to get the team to start thinking about unit testing their code.  You have a couple of functions here that you can easily write some test for, like resolveLabel, and getFormField value.  Even renderElements can be tested with 
import '@testing-library/jest-dom';  Could you please write some tests before merging the code?

libs/jsonforms-components/src/lib/Controls/FormStepper/FormStepperControl.tsx
```diff
@@ -126,6 +134,53 @@ export const FormStepper = ({
     );
   };
 
+  const resolveLabelFromScope = (scope: string) => {
+    const lastSegment = scope.split('/').pop()!;
```
thdlam, line 138: please do you not use .pop()!... This is telling typescript you know when you pop the string array there will be a value.  But this may be incorrect at runtime when the code is being executed.

Use a if statement to check if there is a value when popping the string array and put a default value for 'lastSegment' if it doesnt go into the if statement.

rstyan: Make sure when you write your unit tests that you take this "use case" into account.

libs/jsonforms-components/src/lib/Controls/FormStepper/FormStepper.spec.tsx
```diff
@@ -0,0 +1,52 @@
...
+  });
+});
+describe('getFormFieldValue', () => {
+  it('Should return the correct value for a given scope', () => {
+    const scope1 = '#/properties/firstName';
+    const scope2 = '#/properties/address/properties/street';
+    expect(getFormFieldValue(scope1, data)).toEqual('Alex');
+    expect(getFormFieldValue(scope2, data)).toEqual('Springfield');
+  });
+});
```
rstyan, line 52: This is your opportunity to test for bad data as well.  Like, what happens if the scope is malformed e.g. scope="properties.address.properties.street"?  Or, what happens if properties is misspelled? Or, what if scope is null or an empty string?  Unit testing is as much how the code is handling edge cases as it is about good data.  Try to adopt the mindset of a tester - I'm going to do whatever I can to make this code fail!

btw, I could be wrong, but shouldn't the label be 'First name'?  I thought we were using sentence case?

rstyan (changes requested):

rstyan (approved):
This is much better, thanks Rama.

PR #2705, opened 2024-03-13
feat: add the readOnly to jsonform controls

solittlework:


libs/jsonforms-components/src/lib/Controls/Inputs/InputBooleanControl.tsx
```diff
@@ -53,6 +54,7 @@ export const BooleanComponent = ({
       name={`${path}`}
       checked={data}
       onChange={(name: string, checked: boolean, value: string) => {
+        if (readOnly) return;
```
rstyan, line 57: This would be a lot more maintainable if you wrote a wrapper for the handler function; something like:
```
const readOnlyHandler = (uischema, handler) => {
      if (uischema?.options?.componentProps?.readOnly === true) {
          return () => {}
      } else {
        return handler
      }
    }
```
and then used that in the onChangeFunction

rstyan (changes requested):

thdlam (approved):

libs/jsonforms-components/src/lib/Controls/Inputs/InputBooleanControl.tsx
solittlework: @rstyan thank you for your comments. I moved the event function into an independent function for unit test.

PR #2710, opened 2024-03-14
fix: task management app - duplicated callout errors going to a URL with wrong queue id

athena-chen-chen:


solittlework (approved):

apps/task-app/src/app/state/feedback.slice.ts
```diff
@@ -52,7 +52,7 @@ const feedbackSlice = createSlice({
         };
       }
 
-      state.items.push(item);
+      if (state.items.length === 0 || !state.items.some((obj) => obj?.message === item.message)) state.items.push(item);
```
rstyan, line 55: could you please move the condition into a function whose name describes what you are trying to test?  e.g.
```const containsItem = (items, message) => {
  return items.length !== 0 && items.some((obj) => message === item.message
}```
and then your test would be if (!containsItem(state.items) ....etc.

This will make your code much clearer & cleaner.

rstyan (changes requested):

rstyan (approved):
Thanks!

PR #2711, opened 2024-03-14
fix(jsonforms-components): refactor input event handling and add unit tests

thdlam:
Also added readOnly property for input controls that has this property in it.

### DIFF
--- a/libs/jsonforms-components/src/lib/Controls/Inputs/inputControl.spec.ts
+++ b/libs/jsonforms-components/src/lib/Controls/Inputs/inputControl.spec.ts
@@ -14,6 +14,9 @@ import {
   onKeyPressForDateControl,
   onKeyPressForTextControl,
   onKeyPressNumericControl,
+  onChangeForDateControl,
+  onChangeForDateTimeControl,
+  onChangeForNumericControl,
 } from '../../util/inputControlUtils';
 
 const theDate = {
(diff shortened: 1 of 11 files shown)


athena-chen-chen (approved):

athena-chen-chen:
Looks so clean and precise , Thank you!

libs/jsonforms-components/src/lib/util/inputControlUtils.ts
```diff
@@ -18,3 +21,175 @@ export const isRequiredAndHasNoData = (props: ControlProps) => {
...
+    //so the component will not re render.
+    if (data !== +value) {
+      let newValue: string | number = '';
+      if (value !== '') {
+        newValue = +value;
+      }
+      handleChange(path, newValue);
+    }
+  }
+};
```
rstyan, line 195: Moving these things into utilities is a great idea; less clutter in the other files.  One suggestion though: there seems to be a fair amount of duplicate code.  One way to manage this is to have base functions that you pass a function into, so something like:
```
const onKeyPress = (props: EventKeyPressControlProps, handler: () => void) => {
  return (props) => {
    const { controlProps, key, value } = props;
    const { path } = controlProps;

    if (isNotKeyPressTabOrShift(key)) {
      handler(path, value);
    }
  }
};
```
and then, for example:
```
const onKeyPressForTimeControl = onKeyPress(props: EventKeyPressControlProps, props.handleChange)
```
The reason its only a suggestion is that I'm not sure if it would clean things up sufficiently to be worth while.

libs/jsonforms-components/src/lib/Controls/Inputs/inputControl.spec.ts
```diff
@@ -135,4 +148,265 @@ describe('input control tests', () => {
...
+      const eventProps = {
+        name: 'age',
+        value: 'ab',
+        controlProps: newProps as ControlProps,
+      };
+      onChangeForNumericControl(eventProps);
+      expect(eventProps.value).not.toMatch(regExNumbers);
+      expect(handleChangeMock.mock.calls.length).toBe(0);
+    });
+  });
```
rstyan, line 411: A minor suggestion: The test should read like a book e.g.
    it ('can handle invalid data')
reads "it can handle invalid data".  Of course, what is it, in this case.  So that's where descriptions come in handy:
description('tests for onChangeForNumericControl', () => {
   it('can handle invalid data') { .... etc. },
   it('can handle valid data') { ... etc. }

rstyan (approved):
Nicely done.  I have minor suggestions.  Do with them as you will.

PR #2713, opened 2024-03-14
feat: add the function to allow tenant admins delete core user idp

solittlework:


apps/tenant-management-webapp/src/app/store/tenant/api.ts
```diff
@@ -44,4 +48,24 @@ export class TenantApi {
     const { data } = await this.http.get<TenantsResponse>(url, { params: { adminEmail } });
     return data.results[0];
   }
+
+  async fetchUserIdByEmail(email: string): Promise<string> {
```
thdlam, line 52: Can you write a test the api call for a successful call and not unsuccessful call (400/500)?

thdlam (changes requested):
please add some unit tests for api calls

thdlam (approved):

PR #2720, opened 2024-03-18
fix: GroupLayout only adheres to componentProps outside the controls …

thdlam:
…schema

### DIFF
--- a/libs/jsonforms-components/src/lib/layouts/GroupControl.spec.ts
+++ b/libs/jsonforms-components/src/lib/layouts/GroupControl.spec.ts
@@ -1,0 +1,38 @@
+import '@testing-library/jest-dom';
+import { GoAGroupControlComponent } from './GroupControl';
+import { ControlElement, Layout, LayoutProps } from '@jsonforms/core';
+import { render } from '@testing-library/react';
+
+describe('Input Group Control tests', () => {
+  const groupUiSchema: Layout = {
+    type: 'Group',
+    elements: [
+      {
+        type: 'Control',
+        options: {
+          componentProps: {},
+        },
+      },
+    ],
+  };
+
+  const staticProps: LayoutProps = {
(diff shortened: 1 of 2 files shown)


solittlework (approved):

libs/jsonforms-components/src/lib/layouts/GroupControl.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+  };
+
+  describe('can create control for GoAGroupControlComponent', () => {
+    it('can create control', () => {
+      const props = { ...staticProps };
+      const component = render(GoAGroupControlComponent(props));
+      expect(component).toBeDefined();
+    });
+  });
+});
```
rstyan, line 38: Your test doesn't really test the new functionality you added though.  Like, you should be able to test that the various attributes of the component are correct.  i.e.
```
<GoAContainer type="info" accent="filled" padding="relaxed">
```
does the component have the right accent, or padding, or type?
You've may have run into the issue of unit tests not properly rendering the component though.  You might have to add the following code:
```
/**
 * VERY IMPORTANT:  Rendering <JsonForms ... /> does not work unless the following
 * mocks are included.
 */
window.matchMedia = jest.fn().mockImplementation((query) => {
  return {
    matches: true,
    media: query,
    onchange: null,
    addListener: jest.fn(),
    removeListener: jest.fn(),
    addEventListener: jest.fn(),
    removeEventListener: jest.fn(),
    dispatchEvent: jest.fn(),
  };
});
```
Now, if you log the _component_ you will see proper HTML, and can use the appropriate test functions, like component.getByText() or there might be the equivalent of  a "getByAttribute" .

rstyan (changes requested):

rstyan (approved):
Nice!

PR #2726, opened 2024-03-19
fix: not show layouts with empty elements in the object list

solittlework:


libs/jsonforms-components/src/lib/Controls/ObjectArray/ObjectListControl.tsx
```diff
@@ -168,6 +168,8 @@ const NonEmptyCellComponent = React.memo(function NonEmptyCellComponent({
       };
     }),
   };
+
+  console.log();
```
thdlam, line 172: remove this console log

PR #2735, opened 2024-03-21
feat: review page fields alignment

RamaChandiniDevarapalli:


thdlam (changes requested):
Is it possible to change some of types you are passing into to renderFormFields  with specific types instead of any.
Also write some unit test for renderFormFields.

libs/jsonforms-components/src/lib/Controls/FormStepper/FormStepperControl.tsx
```diff
@@ -57,28 +60,70 @@ export const getFormFieldValue = (scope: string, data: object) => {
...
   } else {
     return '';
   }
 };
 
-export const renderFormFields = (elements: UISchemaElement[] | (Category | Categorization)[], data: object) =>
+export const renderFormFields = (
+  elements: UISchemaElement[] | (Category | Categorization)[],
+  data: any, // eslint-disable-line @typescript-eslint/no-explicit-any
+  ajv: any, // eslint-disable-line @typescript-eslint/no-explicit-any
```
solittlework, line 76: I think we still know the ajv and requiredFiields type.

solittlework (approved):

libs/jsonforms-components/src/lib/Controls/FormStepper/FormStepperControl.tsx
```diff
@@ -157,6 +226,31 @@ export const FormStepper = ({
       setShowNextBtn(true);
     }
   }
+  function getAllRequiredFields() {
+    const requiredFields: string[] = [];
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    function findRequired(fields: any) {
```
rstyan, line 233: Is it really _any_?  Maybe in this case it is, but I see you are using _any_ a lot.  Please try to determine types where possible. Types are your friend.

libs/jsonforms-components/src/lib/Controls/FormStepper/FormStepperControl.tsx
```diff
@@ -157,6 +226,31 @@ export const FormStepper = ({
...
+      } else if (fields.type === 'array' && fields.items && typeof fields.items === 'object') {
+        findRequired(fields.items);
+      }
+    }
+
+    findRequired(schema);
+
+    return requiredFields;
+  }
+
```
rstyan, line 253: Please move this function into a separate file.  Then you function's signature woudl be something like:
```
getAllRequiredFields(fields: Array<something>, schema: JsonSchema): Array<string> { ... }
```
Then, add another file for unit tests on this function.

libs/jsonforms-components/src/lib/Controls/FormStepper/FormStepperControl.tsx
```diff
@@ -57,28 +60,70 @@ export const getFormFieldValue = (scope: string, data: object) => {
...
+      : typeof currentValue === 'object'
+      ? ''
+      : currentValue;
   } else {
     return '';
   }
 };
 
-export const renderFormFields = (elements: UISchemaElement[] | (Category | Categorization)[], data: object) =>
+export const renderFormFields = (
```
rstyan, line 73: Any time you see large functions like this inside a file, please try to extract it into its own, testable, reusable file.  Start thinking about the size of files - generally they should not be larger than a couple of hundred lines of code, at most.  Moving chunks of complex code into separate files makes things much easier to digest for people reviewing, and maintaining this code.  Do you know about separation of concerns?

This function looks quite complex, and should have a large number of unit tests to cover the various bits of logic.  Thats much easier to do if the function can be isolated.

rstyan (changes requested):

rstyan (approved):
Well done, Rama!

thdlam (approved):

PR #2757, opened 2024-03-27
feat(form-app): enable file upload capability for supporting documents

thdlam:


jonathanweyermann (approved):

rstyan (approved):
I see you removed some unit tests, but have not written any new ones. Is there a reason for this?

Otherwise LGTM.

thdlam:
@rstyan I actually didnt remove any tests.  I added quite a bit of jest unit tests for the form.slice.ts Redux file.
I was thinking about writing some unit tests for  FileUploadControl.ts, but didnt had a good idea how i can do this properly to meet code coverage. So I left it out for now.

rstyan:
Oh, sorry I mistook a "load large file" for a "deleted file".
Don't stop unit testing just because you can't do an 80% coverage.  Even 10% is much better than nothing.  Do what you can and defer the rest.  But I'm fine if you want to merge this code.

libs/jsonforms-components/src/lib/Controls/FileUploader/FileUploaderControl.spec.tsx
```diff
@@ -0,0 +1,51 @@
...
+    >
+      <FileUploader {...props} /> ;
+    </ContextProvider>
+  );
+};
+
+describe('FileUploaderControl tests', () => {
+  it('can render fileupload control', () => {
+    const props = { ...staticProps };
+    const { container } = render(contextComponent(props));
```
rstyan, line 47: I ran into a bunch of trouble rendering like this, and it only got resolved when I changed it to:
render(<contextComponent {...props}/>);
It seems like it only rears its ugly head when there are "useState(...)" calls in the component, but its best to keep things consistent.

Also, the React standard is to capitalize the component name, so in this case you would call it "ContextComponent".

thdlam: ok  i fixed this and added one more unit test.

PR #2774, opened 2024-04-02
feature(CS-2969) - submit a form from the stepper

jonathanweyermann:


apps/form-app/src/app/containers/Form.tsx
```diff
@@ -77,6 +77,13 @@ const FormComponent: FunctionComponent<FormProps> = ({ className }) => {
                   showSubmit={showSubmit}
                   saving={busy.saving}
                   onChange={({ data, errors }) => {
+                    if (
+                      errors[0]?.message === 'should be equal to one of the allowed values' &&
+                      errors[0]?.schemaPath.includes('enum')
+                    ) {
+                      errors = null;
+                    }
+
```
rstyan, line 86: This seems a bit dodgy.  Why is this code trying to filter out very specific error messages?  And even if this is a valid thing to do, the code would be much cleaner (and easier to digest) if you made this into a function, e.g.
errors = filterIgnoredErrors(errors)

libs/core-common/src/jsonSchema/ajv.ts
```diff
@@ -21,7 +21,18 @@ export class AjvValidationService implements ValidationService {
...
+      const newSchema = JSON.parse(JSON.stringify(schema));
+
+      Object.keys(newSchema?.properties || {}).forEach((propertyName) => {
+        const property = newSchema.properties || {};
+        if (property[propertyName]?.enum?.length === 1 && property[propertyName]?.enum[0] === '')
+          delete property[propertyName].enum;
+        if (property[propertyName]?.format === 'file-urn') {
+          delete property[propertyName].format;
+        }
+      });
```
rstyan, line 33: This code should also be turned into a function and unit tested separately.

libs/jsonforms-components/src/lib/Context/index.tsx
```diff
@@ -89,12 +89,18 @@ interface FileManagement {
   downloadFile?: (file: File) => void;
   deleteFile?: (file: File) => void;
 }
+interface SubmitManagement {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  submitForm?: (any: any) => void;
```
rstyan, line 94: are the types really any?  There are way too many places in our code where "any" can be easily replaced with an actual type.  This makes the code much, much more managable.

libs/jsonforms-components/src/lib/Controls/FormStepper/FormStepper.spec.tsx
```diff
@@ -0,0 +1,353 @@
...
+          scope: '#/properties/vegetarianOptions/properties/vegan',
+        },
+        {
+          type: 'Control',
+          scope: '#/properties/vegetarianOptions/properties/favoriteVegetable',
+        },
+      ],
+    },
+  ],
+};
```
rstyan, line 308: Again, your tests do not need such elaborate schema's.  Either make them simpler or have you tests use all of the fields.  The way you have it makes it very hard to determine what exactly is being tested and requires the reader to spend too much time figuring out that most of it is unnecessary.

rstyan (changes requested):

apps/form-app/src/app/containers/Form.tsx
jonathanweyermann: I could instead change the ui schema before it gets validated. not sure which is better. maybe that's better

rstyan: Generally speaking, depending on the text of an error message eventually leads to unexpected and hard to detect runtime errors.  If you change the UI schema before had, can you avoid this situation?

jonathanweyermann: yes, I already kind of do a similar thing in this code already, so it wouldn't be hard

rstyan (approved):

PR #2795, opened 2024-04-05
fix(form-app): file upload control ability to upload and delete files 

thdlam:


libs/jsonforms-components/src/lib/common/Ajv.ts
```diff
@@ -2,5 +2,4 @@ import Ajv from 'ajv';
 
 export const ajv = new Ajv({ allErrors: true, verbose: true });
 
-//Example format: urn:ads:platform:file-service:v1:/files/f6de737e-c5fc-42fe-963b-28bfe14597c4
-ajv.addFormat('file-urn', /^urn:ads:platform:file-service:v[0-9]:\/files\/[a-zA-Z0-9.-]*$/);
+ajv.addFormat('file-urn', /^urn:[a-zA-Z0-9.-]+(:[a-zA-Z0-9.-]+)*$/);
```
jonathanweyermann, line 5: don't change this back - someone changed it to 

/^urn:ads:platform:file-service:v[0-9]:\/files\/[a-zA-Z0-9.-]*$/ 

and it's admittedly better - I don't think one you're replacing it with works

libs/core-common/src/jsonSchema/ajv.ts
```diff
@@ -8,7 +8,7 @@ export class AjvValidationService implements ValidationService {
   protected ajv: Ajv.Ajv = new Ajv();
 
   constructor(private logger: Logger) {
-    this.ajv.addFormat('file-urn', /^urn:ads:platform:file-service:v[0-9]:\/files\/[a-zA-Z0-9.-]*$/);
+    this.ajv.addFormat('file-urn', /^urn:[a-zA-Z0-9.-]+(:[a-zA-Z0-9.-]+)*$/);
```
jonathanweyermann, line 11: don't change this back - someone changed it to

/^urn:ads:platform:file-service:v[0-9]:/files/[a-zA-Z0-9.-]*$/

and it's admittedly better - I don't think one you're replacing it with works

jonathanweyermann (changes requested):
what does it fix - also don't change these files back

PR #2842, opened 2024-04-17
fix(resource contention problem) - context provider functions are no …

jonathanweyermann:
…longer global


rstyan (approved):
LGTM

solittlework (dismissed):
Hi @jonathanweyermann, I personally cannot figure out why we call this is react context.
Functional code example: https://react.dev/learn/scaling-up-with-reducer-and-context
Class base example: https://www.taniarascia.com/using-context-api-in-react/

PR #2888, opened 2024-05-01
fix(form-service): add ability to summarize complex objects

rstyan:


libs/jsonforms-components/src/lib/Controls/FormStepper/util/GenerateFormFields.ts
```diff
@@ -1,35 +1,107 @@
...
+const isNestedValue = (value: NestedStringArray): boolean => {
+  return Array.isArray(value) && value.length === 2 && typeof value[0] === 'string' && Array.isArray(value[1]);
+};
+
+/*
+ * Convert ['name', [['first', 'fred'], ['middle', 'jolly'], ['last', 'mcguire']]]
+ * into [['first', 'fred'], ['middle', 'jolly'], ['last', 'mcguire']]
+ */
+const flatten = (arr: NestedStringArray): string[][] => {
+  console.log('flattening');
```
athena-chen-chen, line 67: Please remove debugging code.

libs/jsonforms-components/src/lib/Controls/FormStepper/util/GenerateFormFields.ts
```diff
@@ -1,35 +1,107 @@
-export const resolveLabelFromScope = (scope: string) => {
+import { LabelDescription } from '@jsonforms/core';
+
+type jsonformsLabel = string | boolean | LabelDescription | undefined;
+export const labelToString = (label: jsonformsLabel, scope: string): string => {
+  if (!label) {
+    return resolveLabelFromScope(scope);
+  }
+  if (typeof label === 'object' && label !== null && 'show' in label && label.show) {
```
athena-chen-chen, line 8: if (typeof label === 'object' && label?.show)  is good enough?

libs/jsonforms-components/src/lib/Controls/FormStepper/util/GenerateFormFields.ts
thdlam, line 67: remove these console log/ console.dir

solittlework: Hi @rstyan, just a kind reminder that the jsonform/core includes the lodash lib, which means you can use the lodash flatten, isArray, get or other operations without enlarging the lib size.

libs/jsonforms-components/src/lib/Controls/FormStepper/util/GenerateFormFields.ts
rstyan: No.  On a general object, you first have to test if 'show' is actually in the object, other wise you will get a compile-time error e.g _label?.show_ will have an error something like: "show is not a property of label".  label?.show is a run-time test,  checking to see if label is null before accessing its (known) field.

solittlework (approved):

PR #2906, opened 2024-05-07
fix(status-service) - Accessing configuration too frequently

jonathanweyermann:


apps/status-service/src/app/router/serviceStatus.ts
```diff
@@ -207,7 +207,10 @@ export const updateApplicationStatus =
       const user = req.user as User;
       const { appKey } = req.params;
       const { status } = req.body;
+      console.log(JSON.stringify(status) + '<xx status');
```
thdlam, line 210: just remove this console log

apps/status-service/src/app/router/serviceStatus.ts
```diff
@@ -207,7 +207,10 @@ export const updateApplicationStatus =
...
       const apps = await applicationRepo.getTenantApps(user.tenantId);
+
+      console.log(JSON.stringify(apps) + '<xx apps');
```
thdlam, line 213: this console log as well

thdlam (approved):

apps/status-service/src/app/router/serviceStatus.ts
jonathanweyermann: Thank you

PR #2914, opened 2024-05-09
fix: submit button confirmation modal

RamaChandiniDevarapalli:


jonathanweyermann (changes requested):
Are you sure this works in the form app?

jonathanweyermann (approved):

PR #2930, opened 2024-05-15
feat(form link renderer): Can add Json Forms Link renderer

jonathanweyermann:


libs/jsonforms-components/src/lib/Controls/Link/LinkControl.tsx
```diff
@@ -0,0 +1,76 @@
...
+      </div>
+    </GoAFormItem>
+  );
+};
+
+export const linkControl = (props: ControlProps & WithOptionLabel & TranslateProps) => {
+  return <LinkSelect {...props} />;
+};
+
+export const GoALinkControlTester: RankedTester = rankWith(2, scopeEndsWith('link'));
```
rstyan, line 76: I don't think this is a very good way to distinguish link's from other components.  Like, what if a form contains several links?  They will all have to have the same property name. Surely there is a better way, such as creating a new control type call Link.

jonathanweyermann: ok. I will fix it

PR #2950, opened 2024-05-18
e2e - filter test files by type

howard-zhang:
e2e - filter test files by type


apps/tenant-management-webapp-e2e/src/integration/file-service/file-service.step.ts
```diff
@@ -975,3 +975,70 @@ Then('the user views {string} in security classification dropdown', function (dr
...
+  fileServiceObj
+    .uploadedFilesFilterFileTypeDropdown()
+    .shadow()
+    .find('li')
+    .contains(dropdownItemName)
+    .click({ force: true });
+  cy.wait(1000);
+});
+
+When(
```
LanaDIO, line 989: Formatting, When(

LanaDIO (commented):
Just a small formatting

apps/tenant-management-webapp-e2e/src/integration/file-service/file-service.step.ts
howard-zhang: This is auto-formatted by the following package and we don't want to manual override the default format (which will cause un-necessary future changes and make PR review harder). When the step title is long, the formatter may wrap it to the next line. It happens to this line when the title is long and the title is wrapped into next line.
 "[typescript]": {
    "editor.defaultFormatter": "esbenp.prettier-vscode"
  },

PR #3064, opened 2024-06-11
feat: update jsonform stepper review page

solittlework:


athena-chen-chen (approved):

libs/jsonforms-components/src/lib/Controls/Inputs/InputBooleanControl.tsx
```diff
@@ -23,6 +23,10 @@ export const BooleanComponent = ({
   } as ControlProps);
 
   const text = label + (required ? ' (required)' : '');
+
```
thdlam, line 26: @solittlework can we refactor this section of code to check if isStepperReview.  It looks like some of these controls are the same?

libs/jsonforms-components/src/lib/Controls/Inputs/InputEnumRadios.tsx
```diff
@@ -12,10 +12,15 @@ import { checkFieldValidity } from '../../util/stringUtils';
...
 
 export const RadioGroup = (props: RadioGroupProp): JSX.Element => {
-  const { data, className, id, enabled, schema, uischema, path, handleChange, options, config, label, t } = props;
+  const { data, id, enabled, schema, uischema, path, handleChange, options, config, label, t } = props;
   const enumData = schema?.enum || [];
   const appliedUiSchemaOptions = merge({}, config, props.uischema.options, options);
   const errorsFormInput = checkFieldValidity(props as ControlProps);
+
+  if (uischema?.options?.isStepperReview) {
+    return <div data-testid="input-enum-radios-review">{data}</div>;
```
thdlam, line 21: For these data test ids can you make them unique as there maybe key/ids collisions and it may cause issues with e2e tests

libs/jsonforms-components/src/lib/Controls/Inputs/InputTextControl.tsx
```diff
@@ -20,6 +20,10 @@ export const GoAInputText = (props: GoAInputTextProps): JSX.Element => {
     uischema?.options?.componentProps?.autoCapitalize === true || uischema?.options?.autoCapitalize === true;
   const readOnly = uischema?.options?.componentProps?.readOnly ?? false;
 
+  if (uischema?.options?.isStepperReview) {
+    return <div data-testid="input-text-control-review">{data}</div>;
```
thdlam, line 24: @solittlework can we refactor this section of code to check if isStepperReview is set and rendering the data.  It looks like some of these controls are the same?

libs/jsonforms-components/src/lib/Controls/Inputs/InputEnumRadios.tsx
solittlework: That is a good idea. I will do that for sure.

libs/jsonforms-components/src/lib/Controls/Inputs/InputTextControl.tsx
solittlework: Yeah, I found the simple one is good for most of the cases. But, for some cases, I expect we need to create the text based on the options. This code just provide a framework or placeholder for the future work.

PR #3099, opened 2024-06-19
feat(tenant-admin-webapp): update fetch roles to modals and editors only when roles is required

thdlam:


solittlework (commented):
I will talk with @thdlam and @howard-zhang about the details of the requirement. Please do not merge it at this moment. Thank you!

solittlework (approved):

PR #3133, opened 2024-06-26
feat: update the review in jsonsforms

solittlework:


rstyan (approved):
Looks good Paul.  There is a console.log in one of the files that should be removed.

libs/jsonforms-components/src/lib/Controls/ObjectArray/ObjectListControl.tsx
```diff
@@ -139,17 +148,20 @@ interface NonEmptyRowComponentProps {
...
-  const { schema, errors, enabled, renderers, cells, rowPath, isValid, uischema } = props;
+  const { schema, errors, enabled, renderers, cells, rowPath, isValid, uischema, isInReview } = props;
   const propNames = getValidColumnProps(schema);
   const propScopes = (uischema as ControlElement)?.scope
     ? propNames.map((name) => {
         return `#/properties/${name}`;
       })
     : [];
 
+  console.log(isInReview);
```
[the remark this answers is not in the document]
solittlework: Hi @rstyan, thank you for your review. I deleted it. I tested the Alberta Farm Fuel and it looks good.

libs/jsonforms-components/src/lib/common/Ajv.ts
```diff
@@ -4,4 +4,3 @@ export const ajv = new Ajv({ allErrors: true, verbose: true, strict: 'log' });
 
 //Example format: urn:ads:platform:file-service:v1:/files/f6de737e-c5fc-42fe-963b-28bfe14597c4
 ajv.addFormat('file-urn', /^urn:ads:platform:file-service:v[0-9]:\/files\/[a-zA-Z0-9.-]*$/);
-addFormats(ajv);
```
thdlam, line 7: @solittlework can you confirm this is really what you want to remove?  this may break the form app because i think this is the Ajv object that the form app uses.

solittlework: Hi @thdlam, that is a good question. The key issue here is that we are not expected to use this ajv instance. The jsonform comes with an internal instance. We might need to find a way to add it to the internal one.

PR #3184, opened 2024-07-12
fix: add calender button,access service scroll fix

RamaChandiniDevarapalli:


thdlam (changes requested):
please try using the saveModal approach similar to the file type, pdf, queue editor.

PR #3187, opened 2024-07-13
feat(List with detail) - navigating between cards

jonathanweyermann:


solittlework (approved):
Expect for the filename consistency of the testing files, the rest are good. Thanks!

PR #3264, opened 2024-08-08
Feat/cs 258 create value definitions storing and retrieving

RamaChandiniDevarapalli:


thdlam (changes requested):
Can you change the css styles to use the design tokens?
And what is this --fs-lg and --fw-bold variable and can you change that to use design tokens?

thdlam (approved):

PR #3309, opened 2024-08-19
fix: space in review page for hidden controls

RamaChandiniDevarapalli:


solittlework (commented):
To hide a specific content in the review, we are not expected to change the code in the stepper form.

We might have issues of the hidden rule in the objectWithList or other specific hidden issue at the given control level.

solittlework (approved):

PR #3336, opened 2024-08-23
fix: status scroll issue

RamaChandiniDevarapalli:


apps/tenant-management-webapp/src/app/pages/admin/services/status/testWebhook.tsx
```diff
@@ -71,69 +71,71 @@ export const TestWebhookModal = (): JSX.Element => {
...
-                <PageIndicator />
-              </div>
-            ) : (
-              showEntries &&
-              (entries ? (
-                <JSONFont>{JSON.stringify(entries[0], null, 2)}</JSONFont>
-              ) : (
-                'No timely response from webhook test server - please try again'
-              ))
+      {isOpen && (
```
solittlework, line 74: Is there any specific reason to add isOpen twice?

solittlework (approved):

apps/tenant-management-webapp/src/app/pages/admin/services/status/testWebhook.tsx
RamaChandiniDevarapalli: updated the changes Paul

PR #3364, opened 2024-08-29
 fix(form-gateway): prevent unauthorized access to download and getFile api endpoints.

thdlam:


apps/form-gateway/src/gateway/router.ts
```diff
@@ -62,10 +101,29 @@ async function getFile(
...
+    formResult?.status === FormStatus.Submitted &&
+    formResult?.submitted !== null &&
+    //Need to check the form gateway roles and form service roles, depending where the request was made.
+    //Whether it was from the form app or directly through the form gateway the roles might be different.
+    roles.find((role) => role.includes(FormServiceRoles.Applicant) || role.includes(ServiceRoles.Applicant))
+  );
+}
+
+export function findFile(fileApiUrl: URL, formApiUrl: URL, tokenProvider: TokenProvider): RequestHandler {
   return async (req, res, next) => {
```
tzuge, line 118: Setting aside the user permissions aspect of this, the two endpoints are just too broad in functionality. The requirement for the gateway is to allow an applicant to download the PDF generated for their form submission. So for example, the endpoint should only take the necessarily parameters (i.e. form ID and submission ID if the form creates submission records). The path does not have to match the file API endpoint path, so for example it could be `/download/:formId/:submissionId` or `/download?formId=:formId&submissionId=:submissionI`. There no need to support a criteria object as a query parameter, and no need to pass the user provided value directly through to the file API.

apps/form-gateway/src/gateway/router.ts
```diff
@@ -45,15 +47,52 @@ export function verifyCaptcha(logger: Logger, RECAPTCHA_SECRET: string, SCORE_TH
...
+  tokenProvider: TokenProvider
+): Promise<SimpleFormResponse> {
+  const token = await tokenProvider.getAccessToken();
+
+  const obj = JSON.parse(criteria);
+
+  //formId is contained in the last item of the split array
+  const formId = obj.recordIdContains ? obj.recordIdContains.split('/').at(-1) : '';
+
+  const targetPath = `${formApiUrl.toString()}form/v1/forms/${formId}`;
```
tzuge, line 63: Look at `DirectoryService.getResourceUrl()`. The directory object returned by the SDK can directly resolve a resource URL for you.

apps/form-gateway/src/gateway/router.ts
```diff
@@ -62,10 +101,29 @@ async function getFile(
...
+export function canAccessFile(formResult: SimpleFormResponse, roles: string[], criteria: string) {
+  if (criteria === '{}') return false;
+  if (formResult === null) return false;
+
+  return (
+    formResult?.status === FormStatus.Submitted &&
+    formResult?.submitted !== null &&
+    //Need to check the form gateway roles and form service roles, depending where the request was made.
+    //Whether it was from the form app or directly through the form gateway the roles might be different.
+    roles.find((role) => role.includes(FormServiceRoles.Applicant) || role.includes(ServiceRoles.Applicant))
```
tzuge, line 113: Role based is still not specific enough since all applicants will have the applicant role and shouldn't be able to download each other's PDFs. This needs to check the `form.createdBy.id` against the `user.id`. Only the user who created the form should be able to download it. Ideally it would be restricted to the user who submitted the form, but that's not currently captured.

apps/form-gateway/src/gateway/router.ts
```diff
@@ -45,15 +47,52 @@ export function verifyCaptcha(logger: Logger, RECAPTCHA_SECRET: string, SCORE_TH
...
 async function getFile(
   fileApiUrl: URL,
   tokenProvider: TokenProvider,
   criteria: string,
   tenantId: AdspId
 ): Promise<string> {
   const token = await tokenProvider.getAccessToken();
+
   const encodedCriteria = encodeURIComponent(criteria);
   const targetPath = fileApiUrl.toString() + '/files' + `?criteria=${encodedCriteria}` + `&tenantId=${tenantId}`;
```
tzuge, line 94: Should also include a `top=1` query param since for a given form urn as recordId, it's possible to have multiple submissions and PDFs over time. When submission urn is the recordId, then only one file is expected.

solittlework (approved):

solittlework:
I personally estimate the PR is ready for a test on dev. We can continue to improve it in case there are more issues there. Thanks!

PR #3416, opened 2024-09-11
fix:changing existing tooltips to design system tooltips CS-3421

SreenathSadupallyGoA:


athena-chen-chen:
if infoCircleWithInlineHelpForFile not be used any more, can we remove it. Thanks!

SreenathSadupallyGoA:
> if infoCircleWithInlineHelpForFile not be used any more, can we remove it. Thanks!

Removed it.

athena-chen-chen (approved):

PR #3462, opened 2024-09-18
bug fix for CS-3531 Console error deleting disposition states of a form definition.

Sankeerth-Mulkalla-goa:
Removing an item from the array does not automatically update the array's length. The item is removed from the disposition array, but using splice ensures both the array and its length are updated correctly


jonathanweyermann (changes requested):
no reason to leave in commented code in this case - github will track changes

Sankeerth-Mulkalla-goa:
> no reason to leave in commented code in this case - github will track changes

Update the PR

jonathanweyermann (approved):

PR #3493, opened 2024-09-25
feat(export/import configuration) : support namespace configurations

jonathanweyermann:
image

### DIFF
--- a/apps/tenant-management-webapp/src/app/pages/admin/services/configuration/export/export.tsx
+++ b/apps/tenant-management-webapp/src/app/pages/admin/services/configuration/export/export.tsx
@@ -14,11 +14,8 @@ import { getConfigurationDefinitions, getConfigurations, ServiceId } from '@stor
 import { PageIndicator } from '@components/Indicator';
 import { ConfigurationExportType, Service } from '@store/configuration/model';
 import { Exports, ChipWrapper } from '../styled-components';
-import { ReactComponent as SmallClose } from '@assets/icons/x.svg';
-import { ReactComponent as Triangle } from '@assets/icons/triangle.svg';
-import { ReactComponent as Rectangle } from '@assets/icons/rectangle.svg';
-import { ReactComponent as InfoCircle } from '@assets/icons/info-circle.svg';
 import { NoPaddingH2 } from '@components/AppHeader';
+import { allNames } from '@store/configuration/model';
 
 import { GoAButton, GoACheckbox, GoAChip, GoAContainer, GoAIcon, GoATooltip } from '@abgov/react-components-new';
 
@@ -51,7 +48,11 @@ export const ConfigurationExport: FunctionComponent = () => {
       Object.keys(sortedConfiguration.namespaces).map((namespace) => {
         //eslint-disable-next-line
         return sortedConfiguration.namespaces[namespace].map((name) => {
-          temp[toServiceKey(namespace, name)] = true;
+          if (!name) {
+            temp[namespace] = true;
+          } else {
+            temp[toServiceKey(namespace, name)] = true;
+          }
           totalExportKeys++;
         });
       });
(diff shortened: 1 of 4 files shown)


apps/tenant-management-webapp/src/app/pages/admin/services/configuration/export/export.tsx
```diff
@@ -237,7 +259,7 @@ export const ConfigurationExport: FunctionComponent = () => {
                                       <GoAChip
                                         key={exp}
                                         deletable={true}
-                                        content={shortName}
+                                        content={shortName || 'all names'}
```
rstyan, line 262: Use the constant for all names

apps/tenant-management-webapp/src/app/store/configuration/sagas.ts
```diff
@@ -131,25 +140,38 @@ export function* fetchConfigurations(action: FetchConfigurationsAction): SagaIte
...
-              c.data['description'] = definitions[key]?.configurationSchema?.description;
-            }
-            return c.data;
-          })
+          configs
+            .map((c) => {
+              if (Array.isArray(c)) {
+                return c;
+              }
+              const key = `${c.data.namespace}:${c.data.name}`;
```
rstyan, line 168: I'm a bit lost here: don't we have to worry about the name===null here?  Shouldn't this use the (new) toServiceKey() function?

apps/tenant-management-webapp/src/app/store/configuration/sagas.ts
```diff
@@ -131,25 +140,38 @@ export function* fetchConfigurations(action: FetchConfigurationsAction): SagaIte
...
-  const urls = getFetchUrls(action.services, configBaseUrl);
-  if (configBaseUrl && token && urls.length > 0) {
+
+  const urlsData = getFetchUrls(action.services, configBaseUrl);
+  if (configBaseUrl && token && urlsData.length > 0) {
     try {
       const configs = yield all(
-        urls.map((url) => call(axios.get, url, { headers: { Authorization: `Bearer ${token}` } }))
+        urlsData.map((urlData) => {
+          if (urlData.namespace) {
```
rstyan, line 149: I don't think this if/else is needed, see above comment.

apps/tenant-management-webapp/src/app/store/configuration/sagas.ts
```diff
@@ -118,6 +104,29 @@ export function* fetchConfigurationDefinitions(_action: FetchConfigurationDefini
   }
 }
 
+async function fetchConfigurationNamespace(urlData: UrlData, token: string) {
```
rstyan, line 107: This should work with both mult-page calls and single page calls.  There shouldn't be any need to have the if statement on line 149.  If not, it shouldn't be hard to make it work for both.

solittlework (approved):

apps/tenant-management-webapp/src/app/store/configuration/sagas.ts
```diff
@@ -118,6 +104,29 @@ export function* fetchConfigurationDefinitions(_action: FetchConfigurationDefini
...
+  let hasMoreData = true;
+
+  while (hasMoreData) {
+    let url = `${urlData.url}?top=10&after=${next}`;
+    const { data } = await axios.get(url, { headers: { Authorization: `Bearer ${token}` } });
+
+    configurations.push(...data.results);
+
+    if (data.page?.next) {
+      url = `/configuration/form-service?top=100&after=${data.page.next}`;
```
rstyan, line 120: Why is this hard coded to the form service?  All our services that have paging have the same interface, so why not use ${urlData.url}?top=...

apps/tenant-management-webapp/src/app/store/configuration/model.ts
```diff
@@ -1,5 +1,12 @@
 import { RegisterData } from '@abgov/jsonforms-components';
 
+export const allNames = 'all names';
+
+export interface UrlData {
+  url?: string;
+  namespace?: boolean;
```
rstyan, line 7: I'm not convinced this boolean is needed.  See comments in the sagas.ts file.

rstyan (changes requested):

PR #3506, opened 2024-09-27
feat: Add getFormSubmission to script functions

rstyan:


apps/script-service/Services/IScriptFunctions.cs
```diff
@@ -7,7 +7,7 @@ interface IScriptFunctions
 
   string? GeneratePdf(string templateId, string filename, object values);
   IDictionary<string, object?>? GetConfiguration(string @namespace, string name);
-  FormDataResult? GetFormData(string formId);
```
jonathanweyermann, line 10: Are you getting rid of an existing functionality here ?

apps/script-service/Services/ScriptFunctions.cs
```diff
@@ -73,17 +73,17 @@ public ScriptFunctions(AdspId tenantId, IServiceDirectory directory, Func<Task<s
     return result?.Configuration;
   }
 
-  public FormDataResult? GetFormData(string formId)
+  public virtual FormSubmissionResult? GetFormSubmission(string formId, string submissionId)
```
jonathanweyermann, line 76: Are you getting rid of an existing functionality here as well?

Also what does virtual do in this case?

jonathanweyermann (changes requested):

apps/script-service/Services/IScriptFunctions.cs
rstyan: Good catch.  Musta been brain dead that day.

apps/script-service/Services/ScriptFunctions.cs
rstyan: No, that was a copy-paste problem.  Another good catch.

rstyan: Actually, it is needed for the override in StubScriptFunctions.

jonathanweyermann (approved):

PR #3528, opened 2024-10-02
fix: CS-3585 Unable to submit feedback when rating is selected last

Sankeerth-Mulkalla-goa:
Add additional condition to enable the send button based on radio button states


libs/adsp-feedback/src/main.ts
```diff
@@ -463,6 +463,9 @@ class AdspFeedback implements AdspFeedbackApi {
     if (this.radio2Ref.value && this.radio2Ref.value.checked === true) {
       this.sendButtonRef.value?.removeAttribute('disabled');
     }
+    if (this.radio2Ref.value?.checked === false && this.radio1Ref.value?.checked === true) {
```
rstyan, line 466: This would be much easier to understand if the variable names reflected what actually was being checked instead of radio1 and radio2.  e.g. name them something like "ratingSelector" and "commentSelector". (I know that wasn't your choice, but "always leave the code better than you found it" is a good maxim to code by).

rstyan (changes requested):

libs/adsp-feedback/src/main.ts
```diff
@@ -463,6 +463,9 @@ class AdspFeedback implements AdspFeedbackApi {
...
+      this.sendButtonRef.value?.removeAttribute('disabled');
+    }
```
rstyan, line 468: Could you please add some unit tests to check your changes?  In general, when working on the code, it is best practice to add unit tests to ensure your changes haven't broken anything. (I know, the original author should have added tests, but lets start doing things properly).

rstyan (changes requested):

libs/adsp-feedback/src/main.ts
Sankeerth-Mulkalla-goa: Changed as "ratingSelector" and "commentSelector" instead of radio1 and radio2, for better understanding of variable names

libs/adsp-feedback/src/main.ts
Sankeerth-Mulkalla-goa: Added new test cases to ensure latest code changes haven't broken any existing functionality.

libs/adsp-feedback/src/main.spec.tsx
```diff
@@ -0,0 +1,94 @@
...
+  it('should disable send button if commentSelector is unchecked', () => {
+    adspFeedback['commentSelector'].value.checked = false;
+    const ratingIndex = 3;
+    adspFeedback['selectRating'](ratingIndex);
+
+    // Check that the send button should remain disabled
+    adspFeedback['sendButtonRef'].value.setAttribute('disabled', 'disabled');
+    expect(adspFeedback['sendButtonRef'].value?.disabled).toBe(true);
+  });
+});
```
rstyan, line 94: Nice!

rstyan (approved):

PR #3548, opened 2024-10-08
fix: SIN validation fix

RamaChandiniDevarapalli:


libs/jsonforms-components/src/lib/util/stringUtils.ts
```diff
@@ -86,12 +86,15 @@ interface extractSchema {
  * @returns error message
  */
 export const checkFieldValidity = (props: ControlProps): string => {
-  const { data, errors: ajvErrors, required, label, uischema, schema } = props;
+  const { errors: ajvErrors, required, label, uischema, schema } = props;
+  const { data } = props;
```
thdlam, line 90: can you just put this in one line like:
 const { errors: ajvErrors, required, label, uischema, schema, data } = props;

thdlam (approved):

PR #3554, opened 2024-10-08
feat: add the anonymous file upload in form gateway

solittlework:


apps/form-gateway/src/gateway/router.ts
```diff
@@ -201,6 +205,34 @@ export function downloadFile(
...
+    try {
+      const fileResourceUrl = new URL(`/file/v1/files`, fileApiUrl);
+
+      const token = await tokenProvider.getAccessToken();
+
+      logger.debug(`Submitting form supporting document from anonymous user`, {
+        context: 'GatewayRouter',
+      });
+
+      const { data: responseData } = await axios.post<{ id: string }>(
```
tzuge, line 219: You might have to use the proxy approach instead of axios for calling to the file service since the form upload is a multi-part form data. Also, using the request body with axios like this likely means the file isn't streamed, which could be a problem for large files. See the `downloadFile` handler.

solittlework: Hi @tzuge, thank you for your review. I updated the code based on downloadFile handler and I also added one simple unit test. Please let me know if you have other concerns. Thanks!

athena-chen-chen (approved):

PR #3601, opened 2024-10-18
Fix- CS- 3580 Fix layout issues on the task app task details view

Sankeerth-Mulkalla-goa:


apps/task-app/src/app/containers/details/styled-components.tsx
```diff
@@ -38,23 +38,27 @@ export const ListWithDetailHeading = styled.h3`
 export const ReviewContainer = styled.div`
   display: flex;
   flex-direction: column;
+  align-items:center;
+  padding-top: 24px !important;
```
thdlam, line 42: can you use design tokens for the 24px
use: var(--goa-space-l);

Sankeerth-Mulkalla-goa: @thdlam  updated the PR

thdlam (approved):

PR #3608, opened 2024-10-21
feat(form service table) - Simple object array

jonathanweyermann:
image
image


solittlework (approved):
Hi @jonathanweyermann, thank you for the PR. There are a few things we might need to improve in the future:
* Now, we have a few hard-coded text in the component. Maybe, we might need to enable user to modify them in the future.
* We might need to set a few handler function as options to avoid empty function.

I am still not quite sure how the data is linked with the internal jsonform data storage. There is a dispatch cell function from the jsonform lib to update the cell data. I'd like to test it in the dev environment. And, we still need to check where there are multiple object array components in same page.

PR #3651, opened 2024-10-29
fix(form-definition):missing partially complete label for required fields

SreenathSadupallyGoA:


athena-chen-chen (approved):

solittlework (commented):
Hi @SreenathSadupallyGoA, thank you for your PR. But, it is hard for me to figure out why we need to skip the code blocks for the unit test in the PR.

SreenathSadupallyGoA:
> Hi @SreenathSadupallyGoA, thank you for your PR. But, it is hard for me to figure out why we need to skip the code blocks for the unit test in the PR.

Hi Paul, all the code ignored is already tested in the unit tests in FormStepperControl.test.tsx covering the lines but it still effecting branch coverage which is weird, so I have ignore some unreachable code to get branch overage up.

thdlam:
@SreenathSadupallyGoA can you try write some tests that can trigger that block of code you ignore.  I think you need to access the shadowRoot to trigger the button click.  Probably need to pass in some data for page.

I dont like ignore some blocks of code like that as well.

SreenathSadupallyGoA:
Discussed the scenario with @solittlework regarding why the setTab functionality is ignored.

- The functionality cannot be reached because GoaFormItem doesn’t generate test IDs/ID, so they cannot be individually clicked and tested to verify if setTab is called.

- @solittlework will create a separate task to conduct R&D on this scenario.

solittlework:
Talked with @SreenathSadupallyGoA. We might have issues of test id in the component. We have create a task in the Jira to handle the unit test issue later: https://goa-dio.atlassian.net/browse/CS-3687

PR #3697, opened 2024-11-12
feat: sin error message fix

RamaChandiniDevarapalli:


libs/jsonforms-components/src/lib/Controls/FormStepper/util/validateData.ts
```diff
@@ -13,6 +14,9 @@ export const validateData = (jsonSchema: JsonSchema, data: unknown, ajv: Ajv): b
     if (property[propertyName]?.format === 'file-urn') {
       delete property[propertyName].format;
     }
+    if (property[propertyName]?.title === 'Social insurance number' && data && data[propertyName] === '') {
```
thdlam, line 17: @RamaChandiniDevarapalli, i see you are using the 'Social insurance number' in a number places now.  Can you create export const variable to use so that you can use that variable instead of the hard coded string for maintainability purposes

thdlam (approved):

PR #3699, opened 2024-11-12
fix: Form summary page , press tab could not go Edit

athena-chen-chen:


thdlam (approved):

libs/jsonforms-components/src/lib/Controls/FormStepper/FormStepperControl.tsx
```diff
@@ -254,7 +254,17 @@ export const FormStepper = (props: CategorizationStepperLayoutRendererProps): JS
                       <ReviewItemSection key={index}>
                         <ReviewItemHeader>
                           <ReviewItemTitle>{categoryLabel}</ReviewItemTitle>
-                          <Anchor onClick={() => setPage(index + 1)} data-testid={testId}>
+                          <Anchor
+                            tabIndex={readOnly ? -1 : 0}
+                            onClick={() => setPage(index + 1)}
+                            data-testid={testId}
+                            onKeyDown={(e) => {
+                              if (!readOnly && (e.key === ' ' || e.key === 'Enter')) {
```
thdlam, line 262: Just a thought, maybe we should refactor create a file with all the keys we use Enter, Tab, Space into its own const variable and export it instead of hard coding it.

I have the same issue and i just put it in the DropDownTypes.ts file

athena-chen-chen: Right, I deeply agree, that will make code neat and easy to read.

PR #3701, opened 2024-11-12
fix: delete model focus fix

RamaChandiniDevarapalli:


libs/jsonforms-components/src/lib/Controls/ObjectArray/DeleteDialog.tsx
```diff
@@ -24,6 +24,17 @@ export const DeleteDialog = React.memo(function DeleteDialog({
   title,
   message,
 }: DeleteDialogProps) {
+  useEffect(() => {
+    if (open) {
+      const deleteBtn = document.querySelector('[data-testId="object-array-confirm-button"]') as HTMLButtonElement;
```
thdlam, line 29: are you sure the data test id will not conflict with other controls because not having a unique value might cause issues with Howard's tests nightly test

thdlam (approved):

PR #3706, opened 2024-11-13
fix: Dropdown set focus when an item is selected.

thdlam:


solittlework (commented):
Hi Thang, I am thinking of whether we can use the ref to track the elements. Using plain JS code to get element might cause issue in the virtual dom.

solittlework (approved):

solittlework:
> Hi Thang, I am thinking of whether we can use the ref to track the elements. Using plain JS code to get element might cause issue in the virtual dom.

I talked with Thang  and we might need to create a spike to handle the issue, since the GoAInput might not forward the ref.

PR #3735, opened 2024-11-19
Fix-3690 Script Editor's role tab forces users to scroll back after s…

Sankeerth-Mulkalla-goa:
…election


athena-chen-chen:
Hi, Sankeerth, I checked out your branch and test local dev, the problem is select few role, it select and auto scroll back to top, it happen in both pane and model that have role table.

PR #3829, opened 2024-12-05
feat: Update Address control label for street address

athena-chen-chen:


libs/jsonforms-components/src/lib/Controls/AddressLookup/AddressLookUpControl.tsx
```diff
@@ -161,7 +161,12 @@ export const AddressLookUpControl = (props: AddressLookUpProps): JSX.Element =>
   return (
     <div>
       {renderHelp()}
-      <GoAFormItem label={label} error={errors?.['addressLine1'] ?? ''} data-testId="form-address-line1">
+      <h3>{label}</h3>
```
thdlam, line 164: Hi @athena-chen-chen, can you inspect this H3 tag to see if it contains any --fs- variables.
I am working fixing/removing these missing styles and using design token for CS-3836 that contain these --fs styles and notice form stepper has a h3 and is applying those styles.  You may want to use the design tokens for it.

PR #3962, opened 2025-01-14
fix: handle long names and IDs by breaking word

SreenathSadupallyGoA:


apps/tenant-management-webapp/src/app/pages/admin/services/form/styled-components.ts
```diff
@@ -87,7 +87,7 @@ export const ConfigFormWrapper = styled.div`
   padding-left: 3px;
   border: solid 1px #dcdcdc;
   border-radius: 3px;
-  height: 7.375rem;
+  height: 8.1rem;
```
solittlework, line 90: Hi @SreenathSadupallyGoA, thank you for your update. But, the 0.1 rem is kind of ambiguity. rem is 16 px in browser and px is the smallest unit.  How about 8 rem or 8.125 rem here?

solittlework (approved):

apps/tenant-management-webapp/src/app/pages/admin/services/form/styled-components.ts
SreenathSadupallyGoA: yes @solittlework make sense.

PR #3973, opened 2025-01-16
Form definitions table width is not centered when we have a long desc…

Sankeerth-Mulkalla-goa:
…ription/definition id/name


apps/tenant-management-webapp/src/app/pages/admin/services/form/styled-components.ts
```diff
@@ -413,10 +413,16 @@ export const SpinnerPaddingSmall = styled.div`
...
 `;
 
-export const TableData = styled.td`
-  width: 33.33%;
+export const TableDataName = styled.td`
+  width: 150px;
 `;
 
+export const TableDataId = styled.td`
+  width: 130px;
```
thdlam, line 421: can you change the px to rem

thdlam (approved):

PR #3993, opened 2025-01-21
fix(form-app):remove redirection to logout if session expires for ano…

SreenathSadupallyGoA:
…nymous forms


solittlework (commented):
Hi @SreenathSadupallyGoA, thank you for your update. It looks we changed the loading condition of the LogoutModal and introduced the detection of the Anonymous change in the LogoutModal. Based on my reading, the LogoutModal will not be loaded if the Anonymous is true. I am not sure why we need to detect the Anonymous in the LogoutModal if the loading condition is controlled. I am wondering is there any specific reason for that?

solittlework (approved):

PR #4019, opened 2025-01-24
fix-CS-3900 Duplicated name validation error on editing a notificatio…

Sankeerth-Mulkalla-goa:
…n type

Removing commented code

### DIFF
--- a/apps/tenant-management-webapp/src/app/pages/admin/services/notifications/addEditNotification/addEditNotification.tsx
+++ b/apps/tenant-management-webapp/src/app/pages/admin/services/notifications/addEditNotification/addEditNotification.tsx
@@ -123,6 +123,10 @@ export const NotificationTypeModalForm: FunctionComponent<NotificationTypeFormPr
       name: type.name,
     };
 
+    if (!isEdit) {
+      validations['duplicated'] = type.name;
+    }
+
     if (!type.channels.includes('email')) {
       type.channels = ['email', ...type.channels];
     }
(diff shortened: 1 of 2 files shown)


apps/tenant-management-webapp/src/app/pages/admin/services/notifications/addEditNotification/addEditNotification.tsx
```diff
@@ -205,15 +202,13 @@ export const NotificationTypeModalForm: FunctionComponent<NotificationTypeFormPr
               };
               if (!isEdit) {
                 validators.remove('name');
-                // validations['duplicated'] = value;
```
thdlam, line 208: i think you need to leave/uncomment this validation on this because we need to validate for duplicates when we create a new notification.

apps/tenant-management-webapp/src/app/pages/admin/services/notifications/addEditNotification/addEditNotification.tsx
```diff
@@ -122,9 +122,6 @@ export const NotificationTypeModalForm: FunctionComponent<NotificationTypeFormPr
     const validations = {
       name: type.name,
     };
-    // if (!isEdit) {
-    //   validations['duplicated'] = type.name;
```
thdlam, line 126: uncomment this we need to validate when we create a new notification

apps/tenant-management-webapp/src/app/pages/admin/services/notifications/addEditNotification/addEditNotification.tsx
```diff
@@ -205,15 +202,13 @@ export const NotificationTypeModalForm: FunctionComponent<NotificationTypeFormPr
...
                 validators.remove('name');
-                // validations['duplicated'] = value;
                 validators.checkAll(validations);
               }
               setType({ ...type, name: value, id: isEdit ? type.id : toKebabName(value) });
             }}
             onBlur={() => {
               validators.checkAll({
                 name: type.name,
-                // duplicated: type.name
```
thdlam, line 216: I think for the onBlur event you need to add this 

`     if (!isEdit) {
                validators.checkAll({ name: type.name, duplicated: type.name });
              } else {
                validators.checkAll({ name: type.name });
              }`
as we want to check if is a duplicate name when we create a new notification

thdlam (approved):

PR #4033, opened 2025-01-28
fix(CS-3897 - Object array validation errors) - add required label

jonathanweyermann:
require should now behave according to jsonform specifications

### DIFF
--- a/libs/jsonforms-components/src/lib/util/stringUtils.ts
+++ b/libs/jsonforms-components/src/lib/util/stringUtils.ts
@@ -87,8 +87,7 @@ interface extractSchema {
  * @returns error message
  */
 export const checkFieldValidity = (props: ControlProps): string => {
-  const { data, errors: ajvErrors, required, label, uischema, schema } = props;
-  const labelToUpdate = getLabelText(uischema.scope, label);
+  const { data, errors: ajvErrors, schema } = props;
   const extraSchema = schema as JsonSchema & extractSchema;
 
   if (extraSchema && data && extraSchema?.title === sinTitle) {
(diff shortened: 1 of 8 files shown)


solittlework (approved):

libs/jsonforms-components/src/lib/util/stringUtils.ts
```diff
@@ -98,19 +97,7 @@ export const checkFieldValidity = (props: ControlProps): string => {
       return extraSchema.errorMessage;
     }
   }
-  if (required) {
```
thdlam, line 101: how are these errors being return to the ui if you remove this?

jonathanweyermann: we're not validating this as required here. According to jsonfroms just setting it to required is not enough to trigger a validation error. We also have to set a minLength: 1. Once you do set a minLength, jsonform will trigger an error, and we will pick it up directly - <GoAFormItem error={error?.message ?? ''} mb

libs/jsonforms-components/src/lib/util/stringUtils.ts
```diff
@@ -98,16 +98,16 @@ export const checkFieldValidity = (props: ControlProps): string => {
...
-
     if (schema) {
       if (isEmptyBoolean(schema, data)) {
-        return `${labelToUpdate} is required`;
+        return `${labelToUpdate} is required x`;
       }
 
       if (isEmptyNumber(schema, data)) {
-        return `${labelToUpdate} is required`;
+        return `${labelToUpdate} is required y`;
```
thdlam, line 110: you have a typo

thdlam (approved):

PR #4082, opened 2025-02-10
fix(pages layout) - can use pages layout in stepper

jonathanweyermann:
image
image

image
image


solittlework (commented):
Hi @jonathanweyermann, thank you for the update. I would be better if we can wrapper the new style into separate control (new file) with a new tester.

thdlam (approved):

PR #4085, opened 2025-02-11
fix: update the stepper cache for editor

solittlework:


jonathanweyermann (approved):
Ok. I tested it and it works

PR #4114, opened 2025-02-20
fix: use value instead of label, empty definitions banner

SreenathSadupallyGoA:


apps/tenant-management-webapp/src/app/pages/admin/services/form/definitions/definitions.tsx
```diff
@@ -197,6 +216,9 @@ export const FormDefinitions = ({
 
       {indicator.show && Object.keys(formDefinitions).length === 0 && <PageIndicator />}
       {!indicator.show && !formDefinitions && renderNoItem('form templates')}
+      {showEmptyBanner && (
+        <GoAHeroBanner heading="">There are no form definitions available for the selected tag.</GoAHeroBanner>
```
thdlam, line 220: Can you just change this to a bold text with some spacing like the feedback when there is no feedback sites instead of using hero banner

thdlam (approved):

PR #4200, opened 2025-03-13
feat: update the form app load test

solittlework:


solittlework:
The PR will not affect production line and I am going to merge for a quick test.

PR #4223, opened 2025-03-24
CS-3621

jonathanweyermann:
image


solittlework (approved):

apps/tenant-management-webapp/src/app/components/styled-components.ts
```diff
@@ -8,6 +8,10 @@ export const MarginAdjustment = styled.h4`
 export const PaddingRem = styled.div`
   padding-top: 1rem;
 `;
+export const Padding = styled.div`
+  padding-top: 1rem;
```
thdlam, line 12: Can you change this to use design tokens before you merge

PR #4225, opened 2025-03-24
fix: add delay after resource tag has been created to fetch tags

thdlam:


apps/tenant-management-webapp/src/app/pages/admin/services/form/definitions/addRemoveResourceTagModal.tsx
```diff
@@ -156,8 +156,12 @@ export const AddRemoveResourceTagModal: FunctionComponent<AddRemoveResourceTagMo
...
 
               if (!tagAlreadyAdded()) {
-                dispatch(clearAllTags());
-                dispatch(fetchAllTags());
+                setTimeout(() => {
+                  dispatch(clearAllTags());
+                }, 300);
+                setTimeout(() => {
+                  dispatch(fetchAllTags());
+                }, 300);
```
athena-chen-chen, line 164: Hi, Thang: in this case might be put dispatch into setInterval,  which used to repeatedly dispatch clearAllTags and fetchAllTags every 300 milliseconds until the interval is cleared.

const intervalId = setInterval(() => {
                    dispatch(clearAllTags());
                   dispatch(fetchAllTags());
                  clearInterval(intervalId);
               }, 300);

PR #4237, opened 2025-03-27
CS-3622

jonathanweyermann:
image


apps/tenant-management-webapp/src/app/store/cache/reducers.ts
```diff
@@ -10,11 +15,19 @@ export const defaultState: CacheState = {
 export default function (state: CacheState = defaultState, action: CacheActionTypes): CacheState {
   switch (action.type) {
     case FETCH_CACHE_DEFINITIONS_SUCCESS_ACTION:
+      console.log(JSON.stringify(action.payload) + '<-action.payloadaction.payload.targets');
```
solittlework, line 18: We might need to remove this.

solittlework (approved):
Exper

PR #4241, opened 2025-03-31
fix(jsonforms-components): Add If/Then condition for data schema to check required fields

thdlam:


libs/jsonforms-components/src/lib/Controls/Inputs/inputTextControl.spec.tsx
```diff
@@ -32,13 +32,18 @@ describe('Input Text Control tests', () => {
   const staticProps: GoAInputTextProps & ControlProps = {
     uischema: textBoxUiSchema,
     schema: {},
-    rootSchema: {},
+    rootSchema: {
```
rstyan, line 35: Thang, could you please add some unit tests for this?

rstyan (changes requested):

libs/jsonforms-components/src/lib/Controls/Inputs/inputTextControl.spec.tsx
```diff
@@ -352,4 +354,42 @@ describe('Input Text Control tests', () => {
...
+      expect(schemaProps?.rootSchema).toBeTruthy();
+    });
+
+    it('Then condition should exist if there is a If condition', () => {
+      const rootSchema = schemaProps?.rootSchema as JsonSchema7;
+      expect(schemaProps).toBeTruthy();
+      expect(rootSchema?.if).toBeTruthy();
+      expect(rootSchema?.then).toBeTruthy();
+    });
+  });
```
rstyan, line 394: This isn't quite complete.  There should be some test that check data against the schema, i.e. showing that if the parent condition is matched, then the child element must have a value.

rstyan (approved):
I've approved, but could you take some time to add more to the unit tests once we have the immediate emergency resolved?

PR #4245, opened 2025-04-01
fix(drag and drop upload in form) - supports multiple files now

jonathanweyermann:
image


solittlework (commented):
Hi Jonathan, thanks for the PR. But, there are a few console.log in the PR.

PR #4261, opened 2025-04-11
Cs 4078 - Subscription Channel now allows phone number

jonathanweyermann:


athena-chen-chen (approved):

apps/notification-service/src/notification/model/type.ts
```diff
@@ -19,6 +19,11 @@ import { SubscriberEntity } from './subscriber';
 import { SubscriptionEntity } from './subscription';
 import { NotificationConfiguration } from '../configuration';
 
+function isValidPhoneNumber(str) {
```
thdlam, line 22: Can you keep the regex pattern consistent?
There is another one that has this one the subscribe app
/^[+]?[(]?[0-9]{3}[)]?[-\s.]?[0-9]{3}[-\s.]?[0-9]{4,6}$/im;

PR #4287, opened 2025-04-16
feat: update the form app for multiple file upload

solittlework:


athena-chen-chen (approved):

apps/form-app/src/app/components/DraftForm.tsx
```diff
@@ -112,15 +112,24 @@ export const DraftForm: FunctionComponent<DraftFormProps> = ({
 
   const uploadFormFile = async (file: File, propertyId: string) => {
     const clonedFiles = { ...files };
+    const propertyIdRoot = propertyId.split('.')[0];
```
thdlam, line 115: what does the propertyId string look like now, since you have to split it to an array?

solittlework: Hi @thdlam, to support mutile file upload, we need to add the index of the file in the list for the path. The code is like a path resolver.

PR #4362, opened 2025-05-15
fix(pdf generation) - fix pdf service pdf generation, fix form modal …

jonathanweyermann:
…banner


solittlework (approved):
Hi @jonathanweyermann. the PR is good, but here is one console.log in the PR. I think we can merge it after we remove the console.log.

PR #4400, opened 2025-05-30
feat: add the min and max to jsonform date control

solittlework:


libs/jsonforms-components/src/lib/Controls/Inputs/InputDateControl.tsx
```diff
@@ -60,6 +54,8 @@ export const GoADateInput = (props: GoAInputDateProps): JSX.Element => {
       testId={appliedUiSchemaOptions?.testId || `${id}-input`}
       disabled={!enabled}
       readonly={readOnly}
+      min={minDate && new Date(minDate)}
```
thdlam, line 57: should you set some sort of default value if they put a invalid date in the schema?

solittlework: Hi Thang, the invalid value will not break the component. We can add the custom validator later.

thdlam (approved):

PR #4476, opened 2025-06-30
fix: Script Service after creating a script add script popup still ap…

athena-chen-chen:
…pears and UI improvement for feedback dashboard

### DIFF
--- a/apps/tenant-management-webapp/src/app/pages/admin/services/script/editor/scriptEditorWrapper.tsx
+++ b/apps/tenant-management-webapp/src/app/pages/admin/services/script/editor/scriptEditorWrapper.tsx
@@ -1,7 +1,7 @@
 import React, { useEffect } from 'react';
 import { FunctionComponent, useState } from 'react';
 import { TabletMessage } from '@components/TabletMessage';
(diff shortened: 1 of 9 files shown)


apps/tenant-management-webapp/src/app/pages/admin/services/script/editor/scriptEditorWrapper.tsx
```diff
@@ -171,18 +170,16 @@ export const ScriptEditorWrapper: FunctionComponent<ScriptEditorWrapperProps> =
...
+        <AddScriptModal
+          open={openAddScript}
+          isNew={true}
+          initialValue={selectedScript}
+          onCancel={() => {
+            reset();
+          }}
+          openEditorOnAdd={openEditorOnAdd}
+          onSave={saveScript}
+        />
```
jonathanweyermann, line 182: Make why was this change made? Is it for esthetics only? - make sure it still functions correctly - the original code would cause AddScriptModel to reset the data whenever it's opened - the new change may (depending on how the modal is written) cause some data to persist when it may not be wanted

jonathanweyermann (changes requested):

jonathanweyermann (approved):

PR #4542, opened 2025-07-28
fix: use context when generating a PDF

rstyan:


jonathanweyermann (approved):
I'll approve, assuming it works for you locally

PR #4643, opened 2025-09-09
fix: update the core list of the resource types

solittlework:


apps/tenant-management-webapp/src/app/pages/admin/services/styled-components.ts
```diff
@@ -84,6 +84,7 @@ export const Buttons = styled.div`
 
 export const ActionIconsDiv = styled.div`
   width: 100%;
+  min-width: 4rem;
```
thdlam, line 87: design token plz

solittlework: Thanks, Thang. It has been fixed.

PR #4646, opened 2025-09-10
fix(tenant-management-webapp): Create separate buttons for data and c…

jonathanweyermann:
…onfiguration
image
image


apps/tenant-management-webapp/src/app/pages/admin/services/configuration/editor/ConfigurationData.tsx
```diff
@@ -16,23 +14,17 @@ import { useDebounce } from '@lib/useDebounce';
...
-export const ConfigurationData = ({
-  dataError,
-  setDataError,
-  configurationData,
-  setConfigurationData,
-}: TemplateEditorProps): JSX.Element => {
+export const ConfigurationData = (): JSX.Element => {
   const dispatch = useDispatch();
   const { id } = useParams<{ id: string }>();
+  const [namespace, name] = id.split(':') || [];
```
solittlework, line 22: Hi @jonathanweyermann  is the id the ADSP id? I am thinking of whether we can use the existing lib to determine the namespace and name.

jonathanweyermann: its just a string with namespace:name

not sure if we have a library for that

PR #4723, opened 2025-10-10
fix: add the access service action list for event selection

solittlework:


apps/tenant-management-webapp/src/app/pages/admin/services/script/editor/scriptEditorEventsTab.tsx
```diff
@@ -149,6 +149,7 @@ export const ScriptEditorEventsTab = ({ script, eventNames, onEditorSave }: Scri
   };
 
   const onEdit = (triggerEvent: ScriptItemTriggerEvent) => {
+    console.log(triggerEvent);
```
thdlam: remove console log

PR #4815, opened 2025-11-26
feat : form preview and submissions CS-4549

SreenathSadupallyGoA:
- Load definitions in table (preview mode)
- Loads submissions for the selected definitions
image
image
image

- We can improve this as we go.

### DIFF
--- a/apps/form-management-app/src/app/components/styled-components.ts
+++ b/apps/form-management-app/src/app/components/styled-components.ts
@@ -1,0 +1,122 @@
+import styled from 'styled-components';
+
+export const BackButtonWrapper = styled.button`
+  display: inline-flex;
+  align-items: center;
+  gap: 8px;
+  background: none;
+  border: none;
+  color: var(--goa-color-interactive);
+  cursor: pointer;
+  padding: 8px 12px;
+  border-radius: 4px;
+  font-size: 14px;
+  text-decoration: none;
+  transition: all 0.2s ease;
+
+  &:hover {
+    background-color: var(--goa-color-interactive-hover);
+    color: var(--goa-color-text-light);
+  }
+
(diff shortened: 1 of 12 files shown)


apps/form-management-app/src/app/state/form/selectors.ts
```diff
@@ -1,4 +1,3 @@
-import { AppState } from '../store';
+// The selectors have been moved to form.slice.ts to keep them close to the state definition.
```
smettu007: revert these, lets keep the selectors seperated from the slice to keep it clean, so we know all the selectors are in one place

apps/form-management-app/src/app/components/styled-components.ts
```diff
@@ -0,0 +1,122 @@
+import styled from 'styled-components';
+
+export const BackButtonWrapper = styled.button`
```
smettu007: create a back button folder and add all files related to the button in one folder. and convert this to scss

apps/form-management-app/src/app/state/form/form.slice.ts
```diff
@@ -1,64 +1,286 @@
...
+          message: err.response?.data?.errorMessage || err.message,
+        });
+      } else {
+        throw err;
+      }
+    }
+  }
+);
+
+export const selectDefinition = createAsyncThunk('form/select-definition', (definitionId: string, { dispatch }) => {
```
smettu007: these 2 actions look like they are used for form preview container and the data for selecting the formId or definition id should be managed within the form preview container state and not in the slice. so i suggest moving this logic to the container level where definitionId or submission Id is selected by the components and the container uses these thunks to load the data and return it to the container directly

PR #5277, opened 2026-04-13
feat:(tenant-web-app): add FormAI feature flag and integrate into form editor components

smettu007:
- added feature flag setup just for FormAI tab in the form editor
- fixed an issue where default feature flag values arent being used when config dosent have the flags defined


solittlework (approved):

rstyan (approved):
I'm not familiar with all that needs to be done here, so its difficult to comment.  But you said its working for dev, so I'm ok with the deployment.

PR #5295, opened 2026-04-16
fix(jsonforms-components): CS-4823-Address component layout breaks: Province/Country overlap on small viewport

Sankeerth-Mulkalla-goa:
Adjust address layout responsiveness to stop Province and Country fields from overlapping on narrow screens. Update the Province/Country grid to use a minimum child width so fields stack when space is limited, and change the province dropdown width to 100% so it fits its container consistently. Also aligns the read-only address view with the same responsive behavior.


Sankeerth-Mulkalla-goa:
image

athena-chen-chen (approved):

PR #5347, opened 2026-04-28
feat: update the container of the admin web app

solittlework:


jonathanweyermann (changes requested):
The whole point was to get ride of the white space to the left and right

Image

This reverts that change.. I don't think i get it

jonathanweyermann:
Also, on at 2560px width, it's 4 cards now, which is probably acceptable

PR #5416, opened 2026-05-20
refactor(clean-code-review): reduce noise and improve review focus

Sankeerth-Mulkalla-goa:
Addresses team feedback from first live deployment:

1. Review once per PR - agent skips re-pushes silently
2. Diff only - reviews changed lines only, not the whole file
3. Top 5 violations - posts highest severity issues only
4. Fewer ERRORs - only 2.13, 2.14, 2.18 remain critical
5. Conservative naming - anti-flip-flop rules added per Roy's feedback


solittlework (approved):

.github/scripts/github-pr-clean-code-review.js
```diff
@@ -140,10 +129,18 @@ RULES:
 2.17 Testable Code: Flag functions that are too large or tightly coupled to be independently testable.
 2.18 No Hidden Side Effects: Flag functions that modify external state unexpectedly or produce outputs not indicated by their name.
 
+REACT PATTERNS — DO NOT FLAG:
+The following are standard React patterns and must never be flagged as violations of any rule:
+- useEffect hooks — cannot be cleanly extracted from a component
```
rstyan: useEffect type functions cannot be extracted, its true, but the code within a useEffect may be a candidate for extraction - e.g. if the code within the block is long AND it can easily be refactored into a private function without too many parameters being passed in.  Maybe add instructions like this instead;
React hook review guidance:

_Do not suggest moving a React hook call itself, such as useEffect, useMemo, useCallback, or useState, into a regular helper file or non-hook function.

Hook calls must remain inside a React component or a custom hook.

For complex useEffect bodies, review the code inside the effect separately:
- If logic is pure and does not depend heavily on component state, props, refs, or dispatch functions, suggest extracting it to a helper function.
- If the effect represents a reusable behaviour with clear inputs and outputs, suggest extracting it to a custom hook.
- If the effect is tightly coupled to many local variables, do not suggest extraction unless you can show a clean function signature.
- If extraction would require passing many variables just to recreate the component’s local scope, prefer suggesting smaller internal helper functions or simplifying the effect in place.

Avoid vague comments like “move this useEffect elsewhere.”_

Also, even if we wanted to, sometimes the code is so tightly coupled to the context that it is impossible to move, so a clause like this should be added:

_Before suggesting extraction, estimate the dependency burden.

If the extracted function would require many parameters, many setters, refs, dispatch functions, or locally scoped callbacks, do not recommend moving it to another file.

Instead, suggest one of:
- split the effect into multiple effects by responsibility,
- create small named helper functions inside the component,
- reduce the number of responsibilities in the effect,
- derive values before the effect,
- move pure transformation logic out, while leaving side-effect orchestration in place.
- Do not suggest extraction unless the extracted code has a clean boundary._

rstyan (changes requested):

rstyan (approved):
LGTM

PR #5429, opened 2026-05-22
feat(CS-4943): add AI unit test agent

Sankeerth-Mulkalla-goa:
## What this does
Adds a local AI-powered unit test agent that generates, reviews, and improves unit tests for source files in the ADSP monorepo. Developers run it from the terminal or VS Code, review the output, and commit what they are happy with. Nothing is automated -the developer stays in full control.

## Files added
- .github/scripts/unit-test-agent.js - core agent script, three modes: generate (new test file), review (audit against source), improve (strengthen existing tests). Uses GitHub Models API (GPT-4o) via the same endpoint as the clean code agent. Reads .unittest.yml and unit-testing.md at runtime.
- .github/agents/unit-testing.md - AI skill document loaded into the GPT-4o prompt at runtime. Covers test file location, naming conventions, mock patterns, clean code rules, coverage threshold, and known limitations.
- .unittest.yml - agent configuration. Sets .spec convention, 80% coverage threshold, colocated test location, supported extensions, and excluded paths.

## How to use
Set GITHUB_TOKEN in your terminal, then:
  node .github/scripts/unit-test-agent.js --mode generate --file <path> --dry-run See the developer guide (unit-test-agent-developer-guide.docx) for full instruction
[unit-test-agent-developer-guide.docx](https://github.com/user-attachments/files/28160496/unit-test-agent-developer-guide.docx)
s.


rstyan (changes requested):
I would like to see a means for giving the agent the Jira ticket that describes why the new code is being written, so that the agent has a behavioural understanding of what the code is supposed to do.

However, this looks like a good start Sankeerth.  We will likely have to fine tune it as we discover issues with the unit tests.  Overall, good job.

rstyan (approved):

PR #5479, opened 2026-06-02
feat: add form definition CRUD endpoints to the form service

solittlework:
Adds form definition CRUD endpoints to the form service, removing the need for callers to interact with configuration-service directly.

New endpoints:

POST /form/v1/definitions — create a definition (returns 409 if ID already exists)
PUT /form/v1/definitions/:definitionId — update a definition
DELETE /form/v1/definitions/:definitionId — delete a definition
All three require the form-service:form-admin role and proxy writes to configuration-service internally.

Refactor: Split router/form.ts into router/definition.ts (definition endpoints) and router/form.ts (form instance/submission endpoints) for maintainability. Existing GET definition endpoints are unchanged.

### DIFF
--- a/apps/form-service/src/form/router/definition.ts
+++ b/apps/form-service/src/form/router/definition.ts
@@ -241,7 +241,10 @@ export function createFormDefinitionRouter({
     '/definitions',
     assertAuthenticatedHandler,
     createValidationHandler(
-      body('id').isString().isLength({ min: 1, max: 50 }),
+      body('id')
+        .isString()
+        .isLength({ min: 1, max: 50 })
+        .matches(/^[a-z0-9-]+$/),
       body('name').isString().isLength({ min: 1 }),
       body('anonymousApply').isBoolean(),
       body('applicantRoles').isArray(),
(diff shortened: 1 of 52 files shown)


apps/form-service/src/form/router/definition.ts
```diff
@@ -0,0 +1,289 @@
...
+      const configurationApiUrl = await directory.getServiceUrl(configurationApiId);
+      const token = await tokenProvider.getAccessToken();
+
+      // Check if definition already exists — return 409 if so.
+      const existingDefinitionResponse = await axios.get<{ latest?: { configuration: FormDefinition } }>(
+        new URL(`v2/configuration/form-service/${definition.id}/latest`, configurationApiUrl).href,
+        {
+          headers: { Authorization: `Bearer ${token}` },
+          params: { tenantId: tenantId?.toString() },
+          validateStatus: (status) => status === 200 || status === 404,
```
thdlam: Can you import `HttpStatusCode` from axios and use it like `status ===HttpStatus.Conflict` 

```
import axios, { AxiosInstance, HttpStatusCode } from 'axios';
```

apps/form-service/src/form/router/definition.ts
```diff
@@ -0,0 +1,289 @@
...
+        },
+      );
+      if (existingDefinitionResponse.status === 200 && existingDefinitionResponse.data?.latest?.configuration) {
+        throw new InvalidOperationError(`Form definition with ID '${definition.id}' already exists.`, {
+          statusCode: 409,
```
thdlam: Same use `HttpStatus` instead of using magic numbers

apps/form-service/src/form/router/definition.ts
solittlework: Hi @thdlam, that is a good point and the issue has been resolved. Thanks!

apps/form-service/src/form/router/definition.ts
solittlework: Hi @thdlam, this has been resolved as well.

thdlam (approved):

PR #5484, opened 2026-06-03
feat: update the apis for the form service

solittlework:


apps/form-service/src/utils.ts
```diff
@@ -3,3 +3,13 @@ import { FormDefinition } from './form';
 export function isFormDefinition(value: unknown): value is FormDefinition {
   return typeof (value as FormDefinition)?.anonymousApply === 'boolean';
 }
+
+export function toKebabCase(str: string): string {
```
thdlam: Hi @solittlework, 

Can you add it to this `adsp-monorepo-dev\libs\adsp-service-sdk\src\utils\naming.ts` and centralize here so every other service and app can use it

There is this one in that folder

```
export const toKebabName = (tenantName: string): string => {
  return tenantName.toLowerCase().replace(/ /g, '-');
};
```
 and then there is another one in the form-management-app that is the same

```
export const toKebabName = (name: string): string => {
  return name
    .toLowerCase()
    .replace(/[^a-z0-9\s-]/g, '') // Remove invalid characters
    .replace(/\s+/g, '-') // Replace spaces with hyphens
    .replace(/-+/g, '-') // Replace multiple hyphens with single hyphen
    .replace(/^-|-$/g, ''); // Remove leading/trailing hyphens
};
```

So we dont have to duplicate these utility function everywhere

solittlework: Hi @thdlam, thank you for your review. I've updated code.

thdlam (approved):

thdlam (approved):

PR #5481, opened 2026-06-03
feat(tenant-management-webapp): pdf ai

jonathanweyermann:
image

https://github.com/user-attachments/assets/b51635fa-2329-47a0-9b0c-b3aa8848e146

PR #5500, opened 2026-06-05
fix(clean-code-review): CS-4987-Clean Code AI Agent skips review on PRs opened with more than 1 commit

Sankeerth-Mulkalla-goa:
The agent was silently skipping every PR opened with more than one commit. The previous guard called pulls.listCommits and returned early when commitsResponse.data.length > 1, intending to skip re-reviews on synchronize pushes. Because most feature branches already have multiple commits at the time the PR is opened, this caused the opened event to skip as well - so no review was ever posted.

Root cause confirmed on PR #5481: the branch had 3 commits when the PR was opened, so the very first run hit the > 1 check and exited with "Skipping review - already reviewed on first push." despite no review having been posted. All subsequent synchronize runs also skipped for the same reason.

Fix: pass github.event.action as GITHUB_EVENT_ACTION into the script and gate on event type rather than commit count. The review now runs on opened and reopened events regardless of how many commits are on the branch, and skips on synchronize as intended.

Also fixes two pre-existing test failures:
- buildSystemPrompt rule-presence tests were receiving undefined as rules content because loadConfig's afterEach(jest.resetAllMocks) cleared the fs mock before buildSystemPrompt tests ran. Added a RULES_MOCK constant and dedicated beforeEach/afterEach to the buildSystemPrompt describe block to isolate its mock setup.
- The "top 5 violations" test had a wrong assertion; processFilesForReview collects all violations and the slice(0, 5) cap is applied later in postReviewSummary. Updated the assertion to toBe(6) and renamed the test to reflect the actual behaviour.

### DIFF
--- a/.github/scripts/github-pr-clean-code-review.js
+++ b/.github/scripts/github-pr-clean-code-review.js
@@ -443,9 +443,9 @@ async function main() {
     return;
   }
 
-  const EVENT_ACTION = process.env.GITHUB_EVENT_ACTION;
-  if (EVENT_ACTION !== 'opened' && EVENT_ACTION !== 'reopened') {
-    console.log('Skipping review — only reviewing on PR open/reopen.');
+  const PR_EVENT_ACTION = process.env.GITHUB_EVENT_ACTION;
+  if (PR_EVENT_ACTION !== 'opened' && PR_EVENT_ACTION !== 'reopened') {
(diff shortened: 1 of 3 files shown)


rstyan (approved):
LGTM

smettu007 (approved):

PR #5504, opened 2026-06-05
fix(agent-service): clarify enableConfigurationInvalidation placement in template

tzuge:
## Summary

`enableConfigurationInvalidation: true` was being placed in the second (platform options) argument of `initializeService` instead of the first (service configuration) argument, breaking the service at runtime.

The integration pattern comment now explicitly states: "Add inside the FIRST argument of initializeService (the service configuration object, NOT the second platform-options argument that contains logLevel etc.)".


tzuge:
Consolidated into #5507.

PR #5505, opened 2026-06-05
test(agent-service): add regression tests for nx-adsp template content

tzuge:
## Summary

Adds regression tests for the five categories of template bugs found during beta testing of the nx-adsp agent:

| Test | Bug it catches |
|------|---------------|
| Inventory check | Accidentally removed template |
| TypeScript syntax check (parse) | Syntax errors in newFiles content |
| No unconditional capabilities destructuring | TS6133 unused eventService / tokenProvider |
| Configuration import check | TS6133 unused {ServiceName}Configuration type |
| enableConfigurationInvalidation placement | Property placed in wrong initializeService argument |

The syntax check substitutes `{placeholder}` tokens with a valid identifier before parsing so TypeScript object literals in the template are not consumed by the substitution regex.

Note: full type-checking of template content (to catch `null` assigned to `string` fields) requires the ADSP SDK types to be present and is out of scope here — those bugs live in the nx-adsp generator `.ts__tmpl__` files rather than in these agent-service templates.

### DIFF
--- a/apps/agent-service/src/agent/tools/nxAdspTemplates.spec.ts
+++ b/apps/agent-service/src/agent/tools/nxAdspTemplates.spec.ts
@@ -1,3 +1,4 @@
+import * as ts from 'typescript';
 import { createListNxAdspTemplatesTool, createGetNxAdspTemplateTool } from './nxAdspTemplates';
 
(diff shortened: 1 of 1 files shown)


tzuge:
Consolidated into #5507.

PR #5522, opened 2026-06-10
fix(auto-populate): use middleware instead

jonathanweyermann:


apps/form-app/src/app/components/DraftForm.tsx
```diff
@@ -59,11 +62,50 @@ export const populateDropdown = (schema, enumerators) => {
   return newSchema as JsonSchema;
 };
 
+const shouldAutoPopulate = (actionType: string) => actionType === INIT || actionType === UPDATE_CORE;
```
thdlam: Can you create enum for the actionType instead of using string to use `INIT, UPDATE_CORE, UPDATE_DATA` for actions, maybe later on you need to add a few more to enum instead of someone guessing what string types can be used

thdlam (approved):

PR #5529, opened 2026-06-12
feat: update the form post api

solittlework:
* POST body accepts only name and description
* id always derived from name via toKebabCase
* anonymousApply, applicantRoles, assessorRoles, clerkRoles, dataSchema hardcoded as server defaults
* FormDefinitionCreate schema added to swagger
* FormDefinition.id marked readOnly in swagger
* Stale default-role descriptions removed from swagger
* Obsolete tests removed, new edge-case tests added

### DIFF
--- a/apps/form-service/src/form/router/form.swagger.yml
+++ b/apps/form-service/src/form/router/form.swagger.yml
@@ -10,8 +10,8 @@ components:
       properties:
         id:
           type: string
-          maxLength: 50
-          description: Unique identifier for the form definition. If not provided, will be auto-generated from the name in kebab-case format.
+          readOnly: true
+          description: Unique identifier for the form definition. Always auto-generated from the name in kebab-case format. Cannot be set by the client.
         name:
           type: string
         description:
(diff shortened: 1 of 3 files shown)


apps/form-service/src/form/router/form.swagger.yml
```diff
@@ -24,12 +24,12 @@ components:
           type: boolean
         applicantRoles:
           type: array
-          description: Roles allowed to submit the form. Defaults to ['urn:ads:platform:form-service:form-applicant'] when not provided.
+          description: Roles allowed to submit the form.
```
thdlam, line 27: remove as it is no longer in the request body

apps/form-service/src/form/router/form.swagger.yml
```diff
@@ -24,12 +24,12 @@ components:
...
           items:
             type: string
         assessorRoles:
           type: array
-          description: Roles allowed to assess submitted forms. Defaults to [] when not provided.
+          description: Roles allowed to assess submitted forms.
```
thdlam, line 32: remove as it is no longer in the request body

thdlam (approved):

PR #5530, opened 2026-06-12
feat(form-service): get all data registers from form service

jonathanweyermann:
image

### DIFF
--- a/apps/form-service/src/form/router/form.ts
+++ b/apps/form-service/src/form/router/form.ts
@@ -49,6 +49,9 @@ import {
   SET_TO_DRAFT_FORM_OPERATION,
 } from './types';
 import { PdfService } from '../pdf';
+import * as HttpStatusCodes from 'http-status-codes';
+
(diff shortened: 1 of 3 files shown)


apps/form-service/src/form/router/form.ts
```diff
@@ -556,6 +575,62 @@ export function updateFormData(logger: Logger, fileService: FileService): Reques
...
+export function findDataRegisters(directory: ServiceDirectory, tokenProvider: TokenProvider): RequestHandler {
+  return async (req, res, next) => {
+    try {
+      const tenantId = req.tenant.id?.toString();
+
+      const configurationApiUrl = await directory.getServiceUrl(configurationApiId);
+      const token = await tokenProvider.getAccessToken();
+
+      const tenantConfigurationResponse = await getConfigurationResource<{
+        latest?: {
```
thdlam, line 587: maybe we should create a interface to hold this information instead of remembering what need to set as the return type information and maintainabilty?

thdlam (approved):

PR #5537, opened 2026-06-15
fix(jsonforms-components): auto-populated names can be removed

jonathanweyermann:


apps/form-app/src/app/components/DraftForm.tsx
```diff
@@ -110,17 +111,88 @@ const createMiddleware = (definition, user?: User) => (state, action, defaultRed
   };
 };
 
+const getRequiredMessage = (propertyName: string, property: Record<string, unknown>) => {
+  const label = typeof property.title === 'string' ? property.title : propertyName;
+  return `${label} is required`;
+};
+
+const schemaKeywords = ['items', 'additionalProperties', 'allOf', 'anyOf', 'oneOf', 'not'] as const;
```
solittlework: Hi @jonathanweyermann, I am sorry I am lost here. We'd like to solve the name populate issue. Why we need to add the minLength to all of the required field. Maybe, it is the solution and I am just curious why they are linked.

jonathanweyermann: its because we're removing minimums from required - ie - we used to make the field just disappear if you erased it, but now we show a blank string - but that means that required isn't automatically triggered unless we add another condition if it's added with no condition

thdlam (approved):

PR #5551, opened 2026-06-17
fix(jsonforms-components): add the global event for the change event

solittlework:
* InputBaseTableReviewControl.tsx — Change button calls onChangeDispatch(stepId, scope) via useReviewContext; logic extracted to named handleChangeClick
* InputBaseTableReviewControl.spec.tsx — Added test verifying onReviewChange fires with correct stepId and scope

### DIFF
--- a/libs/jsonforms-components/src/lib/Context/ReviewRenderContext.tsx
+++ b/libs/jsonforms-components/src/lib/Context/ReviewRenderContext.tsx
@@ -1,0 +1,51 @@
+import { createContext, ReactNode, useCallback, useContext, useMemo, useReducer } from 'react';
+
+export interface ReviewChangeState {
+  stepId: number | undefined;
+  scope: string;
+}
+
+type ReviewChangeAction = { type: 'REVIEW_CHANGE'; payload: ReviewChangeState };
+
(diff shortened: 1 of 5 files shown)


libs/jsonforms-components/src/lib/Context/ReviewRenderContext.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+export interface ReviewChangeState {
+  stepId: number | undefined;
+  scope: string;
+}
+
+type ReviewChangeAction = { type: 'REVIEW_CHANGE'; payload: ReviewChangeState };
+
+function reviewReducer(_state: ReviewChangeState | null, action: ReviewChangeAction): ReviewChangeState | null {
+  if (action.type === 'REVIEW_CHANGE') {
```
thdlam: Can you put `REVIEW_CHANGE` as Type instead of hardcoding it

solittlework: hi @thdlam, thanks for the review. It is updated.

thdlam (approved):

PR #5560, opened 2026-06-19
feat(pdf-service): create PDF template

jonathanweyermann:
image


solittlework (approved):

apps/pdf-service/src/pdf/router/pdf.ts
```diff
@@ -72,6 +76,101 @@ export function getTemplate(templateIn: 'params' | 'body'): RequestHandler {
   };
 }
 
+interface PdfTemplateConfiguration {
```
thdlam: Can you move these interfaces to the types.ts to organize it a bit better

thdlam (approved):

PR #5572, opened 2026-06-23
fix(form-service): resolve 404 on GET and PATCH /registers/:name

SreenathSadupallyGoA:


apps/form-service/src/form/router/register.ts
```diff
@@ -35,22 +44,17 @@ export function getRegister(directory: ServiceDirectory, tokenProvider: TokenPro
...
         },
       );
 
-      if (dataResponse.status === HttpStatusCodes.NOT_FOUND || dataResponse.data?.configuration == null) {
-        throw new NotFoundError('data register', name);
-      }
+      assertRegisterExists(dataResponse.status, name);
 
-      const entries = dataResponse.data.configuration as string[];
+      const entries = (dataResponse.data?.latest?.configuration ?? []) as string[];
```
solittlework: The type looks not correct. We can accept object array as well.

solittlework (approved):

PR #5578, opened 2026-06-23
fix: Namespace dropdown opens automatically when Add definition modal…

athena-chen-chen:
… first opens


apps/tenant-management-webapp/jest.config.ts
```diff
@@ -1,11 +1,5 @@
 /* eslint-disable */
-import { createRequire } from 'node:module';
 import path from 'path';
-import { fileURLToPath } from 'url';
```
thdlam, line 4: @athena-chen-chen i think these lines of code that were removed are causing the build errors.
You need to re add it back.  I had the same issue on my PR

PR #5589, opened 2026-06-29
feat(pdf-service):CS-5054 As tenant admin I would like to give a design to the PDF AI generator

Sankeerth-Mulkalla-goa:
Users can now upload a Word document, PDF, or image directly in the AI tab of the PDF Template Editor and the agent will generate an initial template that matches the visual layout of the reference.

Changes:

- pdf.ts: Restructured agent instructions using the RICECO framework (Role, Instructions, Context, Examples, Constraints, Output). Added a dedicated "Required Flow For Design Reference Input" with overwrite protection the agent stops and asks before replacing an existing template, and does not treat the word "generate" in the upload message as confirmation. Agent now generates new headers and footers from the reference content instead of reusing stale defaults, and clears old CSS by passing additionalStyles as an empty string on full template replacement.

- documentParser.ts: Switched DOCX extraction from extractRawText to mammoth.convertToHtml so heading levels, table structure, bold/italic, and inline colors are preserved in the output sent to the LLM. Embedded diagrams are extracted as base64 images (up to 10) and img tags are replaced with [DIAGRAM_N] positional markers so the agent knows exactly where each diagram belongs. Added Unicode ligature normalization to fix broken fi, fl, ff characters that mammoth emits.

- file.ts: After extracting a DOCX, sends the structured HTML to the LLM followed by the extracted diagram images as numbered vision parts, so the agent can reproduce each diagram at its correct position in the template.

- TemplateEditor.tsx: Added handleAttachmentUpload callback wired to AgentChat so file attachments are uploaded to file service and returned as URNs before being sent to the agent. Wrapped in useCallback with error handling via ErrorNotification.

### DIFF
--- a/apps/agent-service/src/agent/agents/pdf.ts
+++ b/apps/agent-service/src/agent/agents/pdf.ts
@@ -9,6 +9,8 @@ description: `Creates, retrieves, modifies, and stores production-ready PDF temp
 using HTML, CSS, and Handlebars syntax.`,
 
 instructions: `
+## R — Role
+
 You are a PDF template editor agent.
 
 The user is always inside the PDF Template Editor.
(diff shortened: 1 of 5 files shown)


Sankeerth-Mulkalla-goa:
https://github.com/user-attachments/assets/25b2c71d-1de5-4463-8e87-b2964f10113a

smettu007 (approved):

PR #5610, opened 2026-07-07
feat(secret-service): configure OpenShift re-encrypt TLS

SreenathSadupallyGoA:


solittlework (approved):

rstyan (changes requested):
Sreenath, have you been able to test this stuff?  Are there any unit tests you could add? Or does all this have to be deployed first?  Actually, I guess it does.  I will redact my block.

rstyan (approved):

SreenathSadupallyGoA:
> Sreenath, have you been able to test this stuff? Are there any unit tests you could add? Or does all this have to be deployed first? Actually, I guess it does. I will redact my block.

@rstyan Ive tested the local fallback path and confirmed the service still starts without the mounted OpenShift cert files, which keeps local/dev execution working. 

Actually, I’ll add unit coverage for the startup selection logic so we explicitly verify both paths,

PR #5627, opened 2026-07-10
feat(pdf-service):CS-5054 give the PDF AI generator visual design references

Sankeerth-Mulkalla-goa:
The agent previously only received extracted text from uploaded PDFs, so generated templates never matched the reference document's actual look. Now PDF pages are rendered as images (first 10 pages, capped at 4MB) and sent to the agent alongside the text, so it can reproduce layout, fonts, and colors from the reference.

agent-service:
- Update tool rejects saves with broken Handlebars or invalid variables JSON and tells the agent how to retry without losing fields. Empty strings clear a field; omitted fields keep their saved values.
- Attachments include a storage note (file type, anonymous read, retention) so the agent can warn when an embedded file won't render in generated PDFs or will be auto-deleted.
- Instructions updated: ask whether the header/footer should match the reference or keep GoA branding, never bind document URNs into img tags, avoid punctuation/brace patterns that break saves.
- New tests for the update tool, file client, and page rendering.

pdf-service:
- New public pdf-template-assets file type (anonymous read, no retention) so images embedded in templates render at generation time.

tenant-management-webapp:
- Editor tabs sync from the agent's tool result instead of refetching stale configuration after a save.
- Image attachments upload to pdf-template-assets with a fallback to agent-attachments when the type isn't registered; documents stay in the protected type.
- Fixed React DOM warning by making isDenseContent a transient prop.

app-common:
- Images can be pasted directly into the agent chat; pasted files get timestamped names since the clipboard reports a generic "image.png".


Sankeerth-Mulkalla-goa:
https://github.com/user-attachments/assets/241c55a3-9608-4c3d-b5e5-fc5789dcc4a0

smettu007 (approved):

PR #5634, opened 2026-07-13
feat: add the data schema and ui schema to the form definition

solittlework:
Summary
Adds a new endpoint for updating the data and UI schemas of an existing form definition without requiring a full definition replace.

Changes
New endpoint — PUT /form/v1/definitions/:definitionId/schemas

Accepts an optional data-schema and/or ui-schema object in the request body
If either field is omitted, the existing value from the stored definition is preserved
Returns 204 No Content on success
Returns 400 if the request body contains a field that is not a valid object
Returns 401 if the caller is not authenticated
Returns 403 if the caller does not have the form-admin role
Returns 404 if no definition with the given ID exists
Refactoring — two shared helpers extracted to eliminate duplication across createFormDefinition, patchFormDefinition, and updateFormDefinitionSchemas:

fetchExistingDefinition — fetches the current definition from the configuration service and throws NotFoundError on a 404 or empty response
patchConfigurationDefinition — performs the configuration service PATCH and wraps axios errors as InvalidOperationError (400)
Swagger — form.swagger.yml updated with the new path, request body schema, and all response codes.

### DIFF
--- a/apps/form-service/src/form/router/definition.spec.ts
+++ b/apps/form-service/src/form/router/definition.spec.ts
@@ -13,6 +13,7 @@ import {
   getFormDefinitions,
   patchFormDefinition,
   updateFormDefinition,
+  updateFormDefinitionSchemas,
 } from './definition';
 
 jest.mock('axios');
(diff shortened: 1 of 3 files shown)


athena-chen-chen (commented):
please fix the clean rule for fetchExistingDefinition.

thdlam (approved):

PR #5639, opened 2026-07-15
fix: Refactor css in feedback widget

athena-chen-chen:
Refactor the CSS in the feedback widget to make the injected styles valid and reduce CSS collisions with the host app.

### DIFF
--- a/libs/adsp-feedback/src/main.spec.ts
+++ b/libs/adsp-feedback/src/main.spec.ts
@@ -10,16 +10,16 @@ describe('AdspFeedback selectRating', () => {
         <div class="rating" data-index="2"><img class="adsp-fb-rating-icon" src="defaultImage3" /></div>
         <div class="rating" data-index="3"><img class="adsp-fb-rating-icon" src="defaultImage4" /></div>
         <div class="rating" data-index="4"><img class="adsp-fb-rating-icon" src="defaultImage5" /></div>
-        <div class="ratingText" data-index="0"></div>
-        <div class="ratingText" data-index="1"></div>
-        <div class="ratingText" data-index="2"></div>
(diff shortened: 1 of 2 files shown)


solittlework (approved):

libs/adsp-feedback/src/main.spec.ts
```diff
@@ -10,16 +10,16 @@ describe('AdspFeedback selectRating', () => {
...
-        <div class="ratingText" data-index="1"></div>
-        <div class="ratingText" data-index="2"></div>
-        <div class="ratingText" data-index="3"></div>
-        <div class="ratingText" data-index="4"></div>
-        <div class="tooltip-text" data-index="0"></div>
-        <div class="tooltip-text" data-index="1"></div>
-        <div class="tooltip-text" data-index="2"></div>
-        <div class="tooltip-text" data-index="3"></div>
-        <div class="tooltip-text" data-index="4"></div>
+        <div class="adsp-fb-rating-text" data-index="0"></div>
```
thdlam, line 13: Hi @athena-chen-chen,
 maybe when you work on the next story for feedback... instead of hard coding `.adsp-fb` in a lot of the places create a const so you can change in one place. I am not sure if we change it a lot, but just a consideration if this happens quite  frequent in the future just to prevent bugs or missing to change it

thdlam (approved):

libs/adsp-feedback/src/main.spec.ts
athena-chen-chen: Good point. Since this PR was focused on isolating the leaking widget classes, I kept the selector changes direct/mechanical to reduce extra churn. If we keep touching the feedback widget styles, I agree it would be worth extracting the base class/selector names into constants so future renames are safer.

PR #5644, opened 2026-07-17
fix: pdf review re-render issue

solittlework:
image


apps/tenant-management-webapp/src/app/pages/admin/services/pdf/templates/previewEditor/PreviewTemplate.tsx
```diff
@@ -94,94 +257,74 @@ export const PreviewTemplate = ({ channelTitle }: PreviewTemplateProps) => {
 
   useEffect(() => {
     const handleWindowResize = () => {
-      if (window.innerHeight !== windowSize) setWindowSize(window.innerHeight);
+      setWindowSize((currentWindowSize) =>
```
thdlam: Can we use `useWindowDimensions` hook that we already have?

thdlam (approved):

PR #5658, opened 2026-07-22
Upgrade ADSP monorepo to Design systems 2.0

athena-chen-chen:
Upgrade the following apps/libs  .

tennant-admin-webapp

status-app

form-management-app

task-app

subscriber-app

libs/form-editor-common


thdlam (approved):

thdlam:
@athena-chen-chen - i am approving the PR, because it looks pretty standard with the changes.  The only thing was the alias method in the webpack.js file... probably you want to test this this PR for DS 1.0 and DS 2.0 with the form app(DS 1.0) and the sandbox (DS2.0) when it is in DEV/UAT  to make sure it works as the way you want it to.

PR #5691, opened 2026-07-30
fix(tenant-management-webapp): align feedback filter date error

jonathanweyermann:
Feedback service → Feedback: when the start date is after the end date, the error icon and message sat on different horizontal levels. Same defect as the calendar event filter.

image


solittlework (approved):

thdlam:
remember to change this to a conventional commit.

PR #5696, opened 2026-07-31
feat(form-service): Add endpoint for updating submission actions

SreenathSadupallyGoA:


apps/form-service/src/form/router/definition.ts
```diff
@@ -674,5 +748,39 @@ export function createFormDefinitionRouter({
...
+        .optional({ nullable: true })
+        .isString()
+        .matches(/^[^:]+:[^:]+$/),
+      body('dispositionStates').optional().isArray(),
+      body('dispositionStates.*.id').optional().isString(),
+      body('dispositionStates.*.name').exists().isString(),
+      body('dispositionStates.*.description').optional().isString(),
+      body().custom((value) => {
+        const extra = Object.keys(value ?? {}).filter((key) => !allowedSubmissionActionProperties.includes(key));
+        if (extra.length > 0) {
```
thdlam, line 776: I dont think you need do a check on the custom method.  If you didn't define the properties that can be use in the body from above it will an error out.

SreenathSadupallyGoA: I think this custom check is still needed. The express-validator field chains validate the listed fields when present, but they do not reject unknown body properties. This endpoint has an acceptance criterion to return 400 for unsupported additional properties, so the allow-list check enforces that explicitly. It also matches the existing roles endpoint pattern just above.

thdlam (approved):

PR #5721, opened 2026-08-07
chore: remove legacy-peer-deps by upgrading yaml to v2

tzuge:
## Summary

`legacy-peer-deps=true` was added to `.npmrc` during the Nx 23 upgrade to work around one specific peer dep conflict — it was the only thing blocking a clean install.

**Root cause:** the root project declared `yaml@^1.10.3` (v1) as a dev dep, but `vite@8.0.16` (pulled in by `@nx/vite@23.1.1`) requires `yaml@^2.4.2` (v2) as an optional peer dep. npm's strict resolver treats this as an ERESOLVE conflict.

**The only v1-specific code:** `tools/scripts/encodeSecrets/index.ts` used `scalarOptions.null.nullStr = ''` — a yaml v1 API removed in v2. In yaml v2 this is a `stringify` option: `stringify(data, { nullStr: '' })`.

**Changes:**
- `package.json`: `yaml` dev dep `^1.10.3` → `^2.4.2`
- `tools/scripts/encodeSecrets/index.ts`: drop `scalarOptions` import and global mutation; pass `{ nullStr: '' }` directly to `stringify()`
- `.npmrc`: deleted (no longer needed)

`npm install` now resolves cleanly with no `--legacy-peer-deps` flag.

## Test plan

- [ ] `npm install` succeeds without `.npmrc` on a clean checkout
- [ ] CI passes
- [ ] `npm run secrets` (encodeSecrets script) produces the same output for null values

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,0 @@
-legacy-peer-deps=true
--- a/package-lock.json
+++ b/package-lock.json
@@ -313,7 +313,7 @@
         "webpack": "5.109.2",
         "webpack-cli": "^7.0.0",
         "webpack-dev-server": "5.2.6",
-        "yaml": "^1.10.3"
+        "yaml": "^2.4.2"
       }
     },
     "node_modules/@a2a-js/sdk": {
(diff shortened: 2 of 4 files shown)


tzuge:
Merged into #5720.

PR #5734, opened 2026-08-10
fix: inconsistent styles in form-editor

SreenathSadupallyGoA:


libs/form-editor-common/src/lib/definitions/style-components.tsx
```diff
@@ -35,6 +35,14 @@ export const QueueTaskDropdown = styled.div`
   margin-top: 0.25rem;
   margin-bottom: 3rem;
 `;
+
+export const LifecycleLabelRow = styled.div`
+  display: flex;
+  align-items: center;
+  gap: 0.5rem;
```
thdlam: use var(--goa-space-2xs) -- for 0.25rem
use var(--goa-space-xs) -- for 0.5rem

SreenathSadupallyGoA: done

thdlam (approved):

PR #5787, opened 2026-08-24
CS-5305: add secret service overview page to TMW

jonathanweyermann:
image

image

Adds the secret service to the tenant management webapp with an overview-only page.

- New `services/secret` page: "Secret service" heading, single Overview tab with the copy from the ticket, standard helpful-links aside.
- Registered in `featureFlag.ts` with `alpha: true` (Alpha badge in sidebar and dashboard card) and `Secret: false` in `defaultFeaturesVisible`, so it stays hidden until enabled per environment.
- `serviceVariables` now sorts by name, so sidebar/dashboard ordering is enforced rather than incidental. No-op against the current list, which was already alphabetical.

Adding secrets is out of scope, so there is no store wiring and no add button.

Note for whoever enables the flag: `tenant-admin.feature:79` asserts dashboard cards by exact count and index, so turning `Secret` on in an e2e environment needs `"Secret"` added to that list and `'The secret service allows'` to `cardTextArray`, both between Script and SharePoint. Left untouched here since the flag is off.

### DIFF
--- a/apps/tenant-management-webapp/src/app/pages/admin/index.spec.tsx
+++ b/apps/tenant-management-webapp/src/app/pages/admin/index.spec.tsx
@@ -1,0 +1,46 @@
+import React from 'react';
+import { Provider } from 'react-redux';
+import { MemoryRouter } from 'react-router-dom';
+import configureStore from 'redux-mock-store';
+import { render } from '@testing-library/react';
+import '@testing-library/jest-dom';
+import TenantManagement from './index';
+import { HeaderCtx } from '@lib/headerContext';
+
+const mockStore = configureStore([]);
+
+const storeWith = (featureFlags: Record<string, boolean>) =>
+  mockStore({
+    config: { serviceUrls: {}, featureFlags, keycloakApi: {} },
+    tenant: { name: 'test-tenant' },
+    session: {
+      authenticated: true,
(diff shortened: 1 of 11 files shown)


jonathanweyermann:
The five RULE-19 warnings above were from the first review (commit 9a863d09c) and are addressed in the amended commit `37cdde779`; GitHub just re-points stale threads at the new head, so they still render inline. The re-review reports `✅ No issues found`.

Fixed by adding colocated specs rather than suppressions: `featureFlag.spec.ts`, `secret/overview.spec.tsx`, `secret/index.spec.tsx`, and `admin/index.spec.tsx` (which covers the ticket AC — routing to the overview page with the flag on, and no route with it off).

Worth noting for future PRs: `// clean-code-ignore: RULE-19` would not have worked here. `checkMissingTestFiles` in .github/scripts/github-pr-clean-code-review.js only honours the repo-wide `disabled_rules` config and never reads suppression comments — that footer line is generic boilerplate emitted for every rule. It does accept `.spec.` alongside `.test.`, so the repo convention satisfies it.

solittlework (approved):

thdlam:
remember to configure the Configmap for the featureFlag once it is deployed

thdlam (approved):

PR #5816, opened 2026-08-31
feat(form-admin-app): CS-5328 open a response from its row

jonathanweyermann:
Per [CS-5328](https://goa-dio.atlassian.net/browse/CS-5328), the Responses table now works the way people expect a list to work.

- The **Actions** column is gone, along with its per-row Open button
- The row under the pointer is highlighted, and only that row
- Clicking anywhere on a row opens the response — the same navigation the Open button performed

### Notes for review

**Keyboard access.** Removing the Open button removed the only keyboard route into a response, so rows take focus and answer Enter and Space, with an accessible name naming the response. A `<tr>` can't honestly claim `role="button"` without breaking table semantics, so this is a focusable row rather than a real button — worth a look if we want a stronger pattern.

**Controls inside the row.** Tags are managed from within the row, so a click that lands on a tag chip or the add-tag button selects the tag instead of opening the response. Covered by a test.

`ActionsCell` and `ActionsColumnHeader` were only used here, so they are removed along with their sticky-column styling; `SelectableRow` takes their place in the same module.

The skeleton row's column count was one too many for the table even before this change; both it and the load-more row now span the four fixed columns plus the data value columns.

### DIFF
--- a/apps/form-admin-app/src/app/components/ResultsTable.tsx
+++ b/apps/form-admin-app/src/app/components/ResultsTable.tsx
@@ -35,19 +35,17 @@ export const ResultsTable: FunctionComponent<GoabTableProps> = (props) => {
   );
 };
 
-// Heading and cell of the trailing column of row actions. The column sticks to the right edge so
-// that actions are still reachable when the table is scrolled.
-const stickyRight = `
-  position: sticky;
-  right: 0;
-  z-index: 1;
-  border-left: 1px solid var(--goa-color-greyscale-200);
-`;
+// Row of a results table that opens its item when selected. The whole row is the target, so it
+// highlights under the pointer and takes focus so it can be selected from the keyboard.
+export const SelectableRow = styled.tr`
+  cursor: pointer;
 
-export const ActionsColumnHeader = styled.th`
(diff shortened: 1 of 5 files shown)


apps/form-admin-app/src/app/components/ResultsTable.tsx
```diff
@@ -35,19 +35,17 @@ export const ResultsTable: FunctionComponent<GoabTableProps> = (props) => {
...
-  ${stickyRight}
-`;
+  &:hover td {
+    background: var(--goa-color-surface-item-hover);
+  }
 
-export const ActionsCell = styled.td`
-  ${stickyRight}
+  &:focus-visible {
+    outline: 2px solid var(--goa-color-interactive-focus);
```
thdlam: can you change this to `var(--goa-space-3xs)`

jonathanweyermann: Done in 9f5b154b2. Also switched the matching `outline-offset` to `calc(var(--goa-space-3xs) * -1)` so the two stay in step.

thdlam (approved):

PR #5837, opened 2026-09-02
fix(jsonforms-components): CS-5347 report every review Change click to the host

jonathanweyermann:
## Problem

A read-only review summary rendered on a different page from the form could not send the user to the step they clicked Change on — it always landed on the application overview.

`useReviewContext().onChangeDispatch` — the call that fires the host's `onReviewChange` — was made by exactly one renderer, `InputBaseTableReviewControl`. Every other Change button only called `goToPage`, which moves the *summary's own* stepper. On a separate host page that context isn't rendering the form, so the click was a no-op, the host never received `(stepId, scope)`, `getExternalFormPath` was never called, and the host fell back to the bare form URL.

That made it look intermittent: plain text/number/date/enum fields go through the one working renderer, so only composite controls failed, silently. The Mailing Address Change button on the sandbox external-navigation demo was already broken for this reason.

## Changes

- New `useReviewChange` hook — the single path a Change click takes. It moves the stepper *and* reports to the review context, so the second half can't be forgotten by the next renderer that adds a Change button.
- Routed all six call sites through it: `FullNameControlReview`, `FullNameDobReviewControl`, `ContactInfoControlReview`, `AddressLookUpControlReview` (both buttons), `ObjectListControl`, and `InputBaseTableReviewControl`.
- `ObjectListControl` was also dropping the scope, so it couldn't focus the field even in-page — it now passes `uischema.scope`.
- Removed a stray `act` import from `InputBaseTableReviewControl` (test-only API in production code, unused).

## Documentation

New tutorial page `docs/tutorials/form-service/external-navigation.md`, which publishes to the developer guide at https://govalta.github.io/adsp-monorepo/tutorials/form-service/external-navigation.html. It walks through the four pieces a host wires up: discovering targets with `getNavigationTargets`, building a URL with `getExternalFormPath` for the cross-*application* case, handing a `navigationTarget` object across a route switch for the same-application case — the one in this ticket, which had no worked example and so read as unsupported — and receiving both through `ContextProvider` / `ReviewRenderProvider`.

The sandbox external navigation page keeps the live demo and links out to that guide.

## Tests

`useReviewChange.spec.tsx` covers the hook, including reporting with no stepper in the tree (the cross-page case) and scope-only targets. Each of the five newly-wired renderers gained a regression test that renders with **no** stepper context and asserts `onReviewChange` fires.

`jsonforms-components` 99 suites / 1458 tests pass, `sandbox-app` 11 / 77 pass, lint clean (0 errors).

## Note for the reviewer

The AC asks for Storybook documentation. There is no Storybook in this repo — no `.storybook` directory and no `*.stories.*` files — so this puts the written guide in the developer guide and leaves the runnable example in the sandbox app. Say if a real Storybook was intended; that is a much larger piece of work.

Refs CS-5347.

### DIFF
--- a/apps/sandbox-app/src/app/components/services/jsonforms/JsonformsExternalNavigation.tsx
+++ b/apps/sandbox-app/src/app/components/services/jsonforms/JsonformsExternalNavigation.tsx
@@ -46,6 +46,7 @@ import {
 } from './externalNavigationDefinition';
 
 const FORM_APP_ID = 'urn:ads:platform:form-app';
+const HOW_TO_URL = 'https://govalta.github.io/adsp-monorepo/tutorials/form-service/external-navigation.html';
 
 const ContextProvider = ContextProviderFactory();
 
@@ -299,6 +300,16 @@ export const JsonformsExternalNavigation = (): JSX.Element => {
             </ContextProvider>
           </>
         )}
+
+        <h3>How to implement this in your application</h3>
+        <GoabText size="body-m">
+          None of the wiring above is discoverable from the rendered result, so it is written up in the ADSP guide:{' '}
+          <a href={HOW_TO_URL} target="_blank" rel="noopener noreferrer">
+            External navigation
+          </a>
+          . It covers both host shapes — a form in a different application, which is what this page demonstrates, and a
+          form on another page of the same application.
+        </GoabText>
       </GoabContainer>
     </ServiceContainer>
   );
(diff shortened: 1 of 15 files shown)


athena-chen-chen (approved):

thdlam:
Story @jonathanweyermann  when i said document it in Storybook i meant the ADSP Guide https://govalta.github.io/adsp-monorepo/

PR #5844, opened 2026-09-03
fix: CS-5343 pin npm to 11.19.1 to clear ip-address advisories

jonathanweyermann:
## What

The root `ip-address` override is already at `10.3.1`. The advisories that remained came from two copies bundled *inside* the `npm` package, which an `ip-address` override cannot reach:

- `node_modules/npm/node_modules/ip-address` @ 10.2.0 — from `npm@11.19.0`
- `node_modules/semantic-release/node_modules/npm/node_modules/ip-address` @ 10.1.0 — from `semantic-release@24.2.9` → `@semantic-release/npm@12.0.2` → `npm@10.9.9`

Pinning `npm` to `11.19.1` via `overrides` fixes both: 11.19.1 bundles `ip-address@10.5.0`, and the nested `npm@10.9.9` subtree dedupes away entirely.

Note `npm@12.0.2` (current `latest`) still bundles `ip-address@10.2.0`, so 11.19.1 is the correct target here, not 12.x.

## Affected areas

`npm` is a dev-only dependency used solely by the semantic-release tooling. `@semantic-release/npm` shells out to the `npm` CLI (`execa("npm", …, { preferLocal: true })`) for `version`, `pack`, `publish`, and `whoami` — all stable across 10 → 11 — so the release pipeline is the only surface, and no runtime service code is touched.

## Verification

- `npm audit` no longer reports `ip-address` (was high, 4 advisories).
- `npm install` resolves cleanly; installed tree matches the lockfile.
- Lockfile diff is confined to the `npm` subtree — no other package changed.


athena-chen-chen (approved):

jonathanweyermann:
Adding the rationale for the `npm` override, since "pinning npm" reasonably raises an eyebrow and the alternatives are worth ruling out explicitly.

## What this override does and does not affect

The `npm` being pinned is **not the npm that installs our dependencies**. It is the npm CLI vendored as a library inside `@semantic-release/npm`, used to run `npm publish` during a release. Installs are untouched — CI gets its npm from `actions/setup-node` with `node-version: 24`, and this override has no bearing on `npm ci` for anyone.

## Why an override on `ip-address` can't do the job

The advisory copies live inside vendored npm trees, which an override on the leaf package cannot reach. Before this change:

| copy | version |
|---|---|
| `node_modules/ip-address` | `10.3.1` — already covered by the existing override |
| `node_modules/npm/node_modules/ip-address` | `10.2.0` — vulnerable |
| `node_modules/semantic-release/node_modules/npm/node_modules/ip-address` | `10.1.0` — vulnerable |

## Alternatives, measured

Resolved from `main`'s lockfile under npm 11 (matching CI), counting vulnerable `ip-address` copies:

| approach | vulnerable copies |
|---|---|
| `main` as-is | 2 |
| flat override `"@semantic-release/npm": "13.1.5"` | 2 — nested copy survives |
| scoped override `"semantic-release": { "@semantic-release/npm": "13.1.5" }` | 2 — nested copy survives |
| **`"npm": "11.19.1"`** (this PR) | **0** |

npm's `overrides` will not rewrite `semantic-release`'s own nested `@semantic-release/npm@12.0.2`, in either form. Pinning `npm` is the only lever that works short of dropping semantic-release. It resolves `npm@11.19.1` with bundled `ip-address@10.5.0` and removes the nested `npm@10.9.9` tree entirely — that removal is the 2,736-line lockfile deletion.

## Residual risk

`@semantic-release/npm@12.0.2` declares `npm ^10.9.3` and now resolves to `11.19.1`, a major outside its range, which npm's `overrides` suppress silently. That is the legitimate objection to this change.

It is very likely inert: all five `.releaserc.json` files name `@semantic-release/npm` explicitly, which resolves to the hoisted `13.1.5`. The `12.0.2` copy is semantic-release's own bundled default-plugin and should never execute. Worth a second opinion from anyone who knows the release path better.

## Direction of travel favours npm 11

`@semantic-release/npm@13.1.5` implements OIDC trusted publishing (`lib/trusted-publishing/`), which is what `id-token: write` in `delivery-ci.yml` is for, and it depends on `npm ^11.6.2`. `12.0.2` predates trusted publishing and wants `npm ^10.9.3`. Keeping the vendored publish client on npm 11 is a requirement of the release path, not an accident — drifting down would break publishing.

## Lockfile verified

The committed lockfile was generated locally under npm 8. Regenerating under npm 11 and diffing gives 0 package differences, 0 version mismatches, same `lockfileVersion: 3` — so it matches what CI resolves.

PR #5902, opened 2026-09-15
feat: notification type having its own detail page

athena-chen-chen:
Screenshot 2026-09-15 at 5 15 30 PM


solittlework (approved):

apps/tenant-management-webapp/src/app/pages/admin/services/notifications/styled-components.ts
```diff
@@ -238,3 +238,159 @@ export const Anchor = styled.div`
...
+  grid-template-columns: auto minmax(0, 1fr);
+  gap: var(--goa-space-m);
+
+  h2 {
+    margin-top: 0;
+  }
+`;
+
+export const NotificationTypeAccent = styled.div`
+  width: 2px;
```
thdlam, line 268: go through the spacing css and update px to use ds token

apps/tenant-management-webapp/src/app/pages/admin/services/notifications/notifications.tsx
```diff
@@ -28,6 +29,13 @@ export const Notifications: FunctionComponent = () => {
     setActivateEditState(edit);
   };
 
+  const handleTabChange = (index: number) => {
+    if (!environment.production && location.pathname.includes('/types/')) {
```
thdlam: is this suppose to route to a different url if it is non prod?

athena-chen-chen: Yes, this is intentional for the temporary DEV-only notification type redesign.
In DEV, notification type detail pages use a nested route like:
/services/notifications/types/:typeId
If the user opens a detail page, then clicks another tab, the URL still contains /types/:typeId. Without clearing that route, clicking back to the “Notification types” tab renders the detail page again instead of the list.
The !environment.production guard keeps this behavior scoped to the DEV-only redesign. UAT/PROD still use the existing legacy Notification Types page and do not expose the new detail route.

thdlam (approved):

PR #5909, opened 2026-09-16
feat: allow partial updates to calendar events - CS-5417

jonathanweyermann:
## Summary
- PATCH /calendar/v1/calendars/:name/events/:id previously required the same body as event creation (name, start, end)
- Adds a dedicated PATCH validation schema where every mutable event property is optional but still validated when supplied
- POST event creation validation is unchanged

## Test plan
- [x] `nx test calendar-service` passes, added a test covering a partial PATCH body (only `end`) that leaves other fields unchanged
- [x] `tsc` on `tsconfig.spec.json` passes
- [x] `nx lint calendar-service` passes with no new issues

Jira: CS-5417


jonathanweyermann:
Thanks for the review, addressing both points:

**1. Is `name` safe to make optional?**
The `name` I made optional is the *event's* `name` field in the PATCH body — not the calendar's `name` route param (`/calendars/:name/events/:id`), which is a required URL segment and is unaffected by this change. `CalendarEventEntity.update()` (`event.ts:53-55`) only overwrites `this.name` when a new value is supplied:

```ts
if (update.name) {
  this.name = update.name;
}
```

So omitting `name` from a PATCH body leaves the existing event name unchanged — nothing throws and the update still happens. Added a test that runs the actual PATCH validation schema against an empty body and confirms it passes.

**2. Start/end ordering validation**
It exists, just not in the request-body validator. It's enforced in `CalendarEventEntity.update()` (`event.ts:77-79`), which throws `InvalidOperationError` if `end < start` *after* applying whatever fields were supplied. Since the entity already holds the persisted event before the update, this correctly catches the case where a partial update (e.g. only `end`) would put it before the existing `start`. This isn't new behavior introduced here — it already applied to full updates and creation. Added two explicit tests for the partial-update case (only `end` supplied, only `start` supplied) to make sure it's covered, so no new bug ticket needed.

Pushed the added test coverage.

thdlam (approved):

PR #5919, opened 2026-09-18
feat(adsp-components): CS-5424 add ADSP theme infrastructure

jonathanweyermann:
image

## Summary
- Add `adsp-components-core`: the framework-independent `AdspTheme` contract and `AdspThemes.standard`, built from GoA design token CSS custom properties (`--goa-*`)
- Add `adsp-components-react`: `AdspThemeProvider` and `useAdspTheme()`, defaulting to `AdspThemes.standard` when no provider is mounted
- No component-specific theme properties yet - `components: {}` is the extension point for later component tickets (CS-5422, CS-5418)

## Test plan
- [x] `nx test adsp-components-core` passes
- [x] `nx test adsp-components-react` passes
- [x] `nx lint adsp-components-core` / `nx lint adsp-components-react` clean
- [x] `tsc --noEmit` on all four lib/spec tsconfigs


libs/adsp-components-core/src/lib/theme/AdspTheme.ts
```diff
@@ -0,0 +1,96 @@
+// clean-code-ignore: RULE-19 — type-only contract, no runtime logic to unit test. Exercised via
```
thdlam, line 1: Can you organize and move this to its own folder under `lib` called it `theme` under the lib folder.  I am thinking we will have more components in this lib folder in the future

libs/adsp-components-react/src/lib/theme/AdspThemeProvider.tsx
```diff
@@ -0,0 +1,30 @@
+import React, { createContext, useContext } from 'react';
```
thdlam, line 1: Same as the `AdspTheme` move it into its own folder under `lib`

thdlam (approved):

